PR #1, opened 2016-12-04
Clean up addr2line so it is a real library

jonhoo:
Sorry for not breaking this down into smaller commits, but there wasn't a good way of doing that without the intermittent commits being unintuitive, broken, or both. This moves all the logic into the library, provides a documented API (with `#[deny(missing_docs)]`), and extends the binary such that its interface matches that of `addr2line`. It builds cleanly on both stable and nightly.

Next steps would probably be to add support for extracting containing function, not just file + line. I *think* this shouldn't be too painful, but you never know. I would also like to try and find a good way to test this, but I'm not entirely sure how to go about that. Suggestions welcome!

The darkest corners of this PR is the nested use of `owning_ref::OwningHandle`. This is necessary so that we can construct a `struct` which contains a `memmap`, an `object::File` that borrows the map, *and* `gimli` objects that borrow the `object::File`. Luckily, it all works out somewhat nicely. The biggest bummer at the moment is that errors in the construction of the `object::File` and `gimli` objects cause panics, which is due to the lack of being able to forward errors through `owning_ref::OwningHandle`. I've filed [a bug](https://github.com/Kimundi/owning-ref-rs/issues/19) about this, so hopefully it will be resolved in not too long.

I've left the `Unit` stuff mostly intact, including the comment saying
> TODO: most of this should be moved to the main library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.2.0"
 
 [dependencies]
 fallible-iterator = "0.1.3"
-clap = "2.*"
+clap = "2.19.1"
 gimli = "0.9.0"
 memmap = "0.5.0"
 object = "0.1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,14 +7,15 @@ license = "Apache-2.0/MIT"
 name = "addr2line"
 readme = "./README.md"
 repository = "https://github.com/gimli-rs/addr2line"
-version = "0.1.0"
+version = "0.2.0"
 
 [dependencies]
 fallible-iterator = "0.1.3"
-getopts = "0.2.14"
+clap = "2.*"
 gimli = "0.9.0"
 memmap = "0.5.0"
 object = "0.1.0"
+owning_ref = { git = "https://github.com/Kimundi/owning-ref-rs" } # need OwningHandle
 
 [features]
 nightly = []
--- a/src/bin/addr2line.rs
+++ b/src/bin/addr2line.rs
@@ -1,36 +1,90 @@
-extern crate fallible_iterator;
-extern crate gimli;
-extern crate getopts;
-extern crate memmap;
-extern crate object;
+#[macro_use]
+extern crate clap;
+extern crate addr2line;
 
-use fallible_iterator::FallibleIterator;
-use object::Object;
-use std::env;
-use std::fs;
+use clap::{App, Arg};
+
+use std::path;
+
+#[cfg_attr(rustfmt, rustfmt_skip)]
+const POSTSCRIPT: &'static str = "\
+addr2line translates addresses into file names and line numbers.
+Given an address in an executable or an offset in a section of a
+relocatable object, it uses the debugging information to figure
+out which file name and line number are associated with it.
+
+By default, the format of the output is `FILENAME:LINENO', and
+each input address generates one line of output. This behavior
+can be modified by passing the -a option, which also includes
+the address on its own line.
+
+If the file name can not be determined, addr2line will print two
+question marks in their place. If the line number can not be
+determined, addr2line will print 0.
+";
 
 fn main() {
-    let args: Vec<String> = env::args().collect();
-    let mut opts = getopts::Options::new();
-    opts.optopt("e",
-                "exe",
-                "Set the input file name (default is a.out)",
-                "<executable>");
+    let matches = App::new("addr2line")
+        .version(crate_version!())
+        .about("Convert addresses into line number/file name pairs.\n\
+                If no addresses are specified on the command line, they will be read from stdin.")
+        .arg(Arg::with_name("executable")
+            .short("e")
+            .long("exe")
+            .default_value("a.out")
+            .help("Specify the name of the executable for which addresses should be translated.")
+            .takes_value(true))
+        .arg(Arg::with_name("addresses")
+            .short("a")
+            .long("addresses")
+            .help("Display the address before the function name, file and line number \
+                   information. The address is printed with a `0x' prefix to easily identify \
+                   it."))
+        .arg(Arg::with_name("addr")
+            .multiple(true)
+            .index(1))
+        .after_help(POSTSCRIPT)
+        .get_matches();
 
-    let matches = opts.parse(&args[1..]).unwrap();
+    let exe = matches.value_of("executable").unwrap_or("./a.out");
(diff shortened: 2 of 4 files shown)


fitzgen:
Thanks for the pull request! Digging in now...

> Sorry for not breaking this down into smaller commits, but there wasn't a good way of doing that without the intermittent commits being unintuitive, broken, or both.

No worries.

> This moves all the logic into the library, provides a documented API (with #[deny(missing_docs)]), and extends the binary such that its interface matches that of addr2line. It builds cleanly on both stable and nightly.

Thank you!!

> I would also like to try and find a good way to test this, but I'm not entirely sure how to go about that. Suggestions welcome!

It should be easy enough to write integration tests by using `nm` on our `addr2line` itself to find addresses and function names, and then pass those addresses into our `addr2line` and make sure that we get *something*.

We could also try comparing our results with the canonical `addr2line`'s results.

As far as unit tests go, its more about semi-tediously writing test data and breaking down each function as far as possible.

These things can, of course, be done in follow up PRs if you are so inclined :)

Cargo.toml
```diff
@@ -7,14 +7,15 @@ license = "Apache-2.0/MIT"
...
+version = "0.2.0"
 
 [dependencies]
 fallible-iterator = "0.1.3"
-getopts = "0.2.14"
+clap = "2.*"
 gimli = "0.9.0"
 memmap = "0.5.0"
 object = "0.1.0"
+owning_ref = { git = "https://github.com/Kimundi/owning-ref-rs" } # need OwningHandle
```
fitzgen: This is just a placeholder until the next release is cut? We won't be able to publish on crates.io with a git dependency :-/

Cargo.toml
```diff
@@ -7,14 +7,15 @@ license = "Apache-2.0/MIT"
 name = "addr2line"
 readme = "./README.md"
 repository = "https://github.com/gimli-rs/addr2line"
-version = "0.1.0"
+version = "0.2.0"
 
 [dependencies]
 fallible-iterator = "0.1.3"
-getopts = "0.2.14"
+clap = "2.*"
```
fitzgen: I'd prefer specifying a proper version here. Cargo will automatically treat `2.3.6` as `^2.3.6`, which is anything in the range `[2.3.6, 3.0.0)`.

src/bin/addr2line.rs
```diff
@@ -1,240 +1,97 @@
...
+fn main() {
+    let matches = App::new("addr2line")
+        .version(crate_version!())
+        .about("Convert addresses into line number/file name pairs.\n\
+                If no addresses are specified on the command line, they will be read from stdin.")
+        .arg(Arg::with_name("executable")
+            .short("e")
+            .long("exe")
+            .default_value("a.out")
+            .help("Specify the name of the executable for which addresses should be translated.")
```
fitzgen: s/name/path/ ?

Cargo.toml
jonhoo: Correct.

src/bin/addr2line.rs
jonhoo: I took the help text verbatim from the `addr2line` [man page](https://sourceware.org/binutils/docs/binutils/addr2line.html) for consistency, but I'm happy to change it if you prefer path.

jonhoo:
> It should be easy enough to write integration tests by using `nm` on our `addr2line` itself to find addresses and function names, and then pass those addresses into our `addr2line` and make sure that we get something.

Do we have a way of referring to the path to build artifacts from inside an integration test?
Also, how were you thinking of using `nm` here? It wouldn't actually give us the mapping from source line to address (though we could cook something up by using placeholder functions).

> We could also try comparing our results with the canonical `addr2line`'s results.

This was more in line with what I was thinking, though the questions above still apply with regards to getting access to our own build artifact in the tests. Also, can we rely on the test running (Travis) to have `addr2line`?

> As far as unit tests go, its more about semi-tediously writing test data and breaking down each function as far as possible.
>
> These things can, of course, be done in follow up PRs if you are so inclined :)

I think I'd prefer for them to be follow-up PRs :)

Cargo.toml
fitzgen: 2.3.6 was just an example. The latest version is perfect, thanks!

src/bin/addr2line.rs
fitzgen: Ah ok, disregard then.

jonhoo:
Just pushed the `Cargo.toml` version change. Do you also want the commits squashed?

fitzgen:
> Do we have a way of referring to the path to build artifacts from inside an integration test?

From http://doc.crates.io/environment-variables.html:

> `CARGO_TARGET_DIR` - Location of where to place all generated artifacts, relative to the current working directory.

So we could do something like this:

```rust
use std::env;

let mut addr2line = PathBuf::new(env::var("CARGO_TARGET_DIR").unwrap());
addr2line.push("addr2line");
```

> Also, how were you thinking of using nm here? It wouldn't actually give us the mapping from source line to address (though we could cook something up by using placeholder functions).

Right -- we would just want to make sure that we didn't get `?:?`, but I'm not sure how reliable that would actually be... Maybe if we filtered for only known functions within `gimli` or `addr2line`...

> Also, can we rely on the test running (Travis) to have addr2line?

I *think* so. If not, we can install it with `apt` from whatever package it might be in from inside `.travis.yml`.

> I think I'd prefer for them to be follow-up PRs :)

Of course :)

> Do you also want the commits squashed?

Nah, its fine.

Thanks again! Very appreciated!

jonhoo:
Unfortunately `CARGO_TARGET_DIR` is a configuration option, not something that's set by `cargo`. We may be able to use `OUT_DIR`, but I don't know that there's any guarantee that the binary has been built by the time the tests run.. :/

jonhoo:
One alternative is to just stuff some C code in there and use [native code compilation](http://doc.crates.io/build-script.html#case-study-building-some-native-code).

philipc:
What would the C code do that we can't do in rust?

You could look at how others have solved it, eg https://github.com/BurntSushi/ripgrep/blob/master/tests/workdir.rs, doesn't look like there is a simple solution though.

jonhoo:
It's just easier to compile reliably from running commands is all. See https://github.com/gimli-rs/addr2line/pull/4. `std::env::current_exe` is a neat trick that might be applicable, though I think getting up-and-running on analyzing a trivial C program first is a good first step.

PR #4, opened 2016-12-04
Add integration tests

jonhoo:
While not particularly extensive, this PR adds the first `addr2line` integration test. It compares the output of the system `addr2line` program with that of the library, and makes sure that whenever we give affirmative answers, our answers match that of the original `addr2line`. It also demonstrates how to write integration tests in a reasonably straightforward manner that also work on Travis, and actually exercises a surprising fraction of the code

Note that the tests [currently fail on macOS](https://travis-ci.org/jonhoo/addr2line/builds/181077541), apparently because the `debug_info` section is missing from the compiled binary. Neither forcing `clang` nor `-gdwarf-4` made a difference. Given that I don't have a macOS computer to test on, I'd appreciate it if someone who does could try to reproduce by running the tests locally.

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -1,0 +1,173 @@
+extern crate addr2line;
+extern crate itertools;
+extern crate glob;
+
+use std::env;
+use std::path;
+use std::process;
+use std::io::prelude::*;
+use itertools::Itertools;
+
+#[test]
+fn identity_map() {
+    let target = env::current_exe().unwrap();
+    let mut debug = target.clone();
+
+    // On macOS, symbols are kept in a separate .dSYM file
+    if cfg!(target_os = "macos") {
+        let fname = debug.file_name().unwrap().to_string_lossy().into_owned();
+        debug.set_file_name(format!("{}.dSYM", fname));
+        debug.push("Contents");
+        debug.push("Resources");
+        debug.push("DWARF");
+        debug.push(fname);
+    }
+
+    // Parse the debug symbols using "our" addr2line
+    let ours = addr2line::Mapping::new(&debug).unwrap();
+
+    // Spin up the "real" addr2line
+    let mut theirs = spawn_oracle(target.as_path());
+
+    // Find some addresses to test using nm (we'll later filter to just text section symbols)
+    let mut theirs_in = theirs.stdin.take().unwrap();
+    let mappings = get_test_addresses(target.as_path(), &mut theirs_in);
+    drop(theirs_in);
+
+    // Go through addresses one by one, and check that we get the same answer as addr2line does.
+    use std::io::BufReader;
+    let mut theirs_out = BufReader::new(theirs.stdout.take().unwrap()).lines();
+    let mut all = 0;
+    let mut excusable = 0;
+    let mut got = 0;
+    for addr in mappings {
+        // Read the oracle ouput
+        let oracle = theirs_out.next().unwrap().unwrap();
+        let oracle = canonicalize_oracle_output(&*oracle);
+
+        // Does our answer match?
+        if let Some((mut file, lineno)) = ours.locate(addr) {
+            // We dared to guess -- did we give the right answer?
+            if cfg!(target_os = "macos") {
+                // atos doesn't include the full path, just file name
+                let f = path::PathBuf::from(&file);
+                let f = f.file_name().unwrap().to_string_lossy().into_owned();
+                file = f;
+            }
+
+            assert_eq!(oracle.0, Some(&*file));
+            assert_eq!(oracle.1, Some(lineno));
+            got += 1;
+            all += 1;
+        } else {
+            if oracle.0.is_some() {
+                // addr2line found something, and we didn't :(
+                println!("we missed 0x{:08x}: {:?}", addr, oracle);
+                all += 1;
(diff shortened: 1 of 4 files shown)


fitzgen:
Don't have time to read the code right now, cause I'm about to go to bed, but

> Note that the tests currently fail on macOS, apparently because the debug_info section is missing from the compiled binary. Neither forcing clang nor -gdwarf-4 made a difference. Given that I don't have a macOS computer to test on, I'd appreciate it if someone who does could try to reproduce by running the tests locally.

On mac os, the debug info ends up split form the executable, so if you had `path/to/executable`, the debug info is going to be in something like `path/to/executable.dSYM/Contents/Resources/DWARF/executable`. Maybe a little different, that's from memory and I'm in Fedora right now.

Will check this out in more detail tomorrow -- thanks for doing this!

jonhoo:
@fitzgen no such luck with the path modification unfortunately. I'm heading to bed too. It'd be great if you could take a look when you get a chance. Whatever we find to be the solution we'll probably also [want to include](https://github.com/gimli-rs/addr2line/issues/5) in the main `addr2line` command line tool. Come to think of it, we may also need to modify the tests on macOS to use `atos` instead of `addr2line`, which will require some changes.

jonhoo:
@fitzgen finally figured it all out -- phew, that was a pain without a macOS computer to debug on! See inline code comments + commit message for details. Feel free to delete all the leftover coveralls comments (except, obviously, the last one). I've squashed all the debug commits I made, and the commit should now be ready for review.

jonhoo:
Seems I spoke too soon -- macOS passes on stable, but beta and nightly both have an off-by-one on the line number... No idea what the best thing to do here is.

jonhoo:
Also, for some reason, the `kcov` step on `linux-nightly` now times out (even though it didn't in earlier instantiations of the tests). Do you have any experience with this @fitzgen?

fitzgen:
> Also, for some reason, the kcov step on linux-nightly now times out (even though it didn't in earlier instantiations of the tests). Do you have any experience with this @fitzgen?

Interesting. I think @philipc knows a little bit more about `kcov` things than I do -- I've just been copying his `coverage` helper script around to new repos :P

jonhoo:
Fixed that by skipping every fifth address output by `nm` (and thus significantly reducing test runtime).

fitzgen:
> Seems I spoke too soon -- macOS passes on stable, but beta and nightly both have an off-by-one on the line number... No idea what the best thing to do here is.

I think we can take an incremental approach, where we skip the tests on OSX and just ensure it builds, and try to fix the tests in follow ups.

tests/integration_test.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::process;
+use std::io::prelude::*;
+use itertools::Itertools;
+
+#[test]
+fn identity_map() {
+    let target = env::current_exe().unwrap();
+    let mut debug = target.clone();
+
+    // point us at the real macOS debugging files
```
fitzgen: Nitpick: Can you capitalize + punctuate this comment and the others below, so they're proper sentences?

tests/integration_test.rs
```diff
@@ -0,0 +1,133 @@
...
+    // create our mapping
+    let ours = addr2line::Mapping::new(&debug).unwrap();
+
+    // spin up the "real" addr2line
+    let (theirs, exe) = if cfg!(target_os = "macos") {
+        ("atos", "-o")
+    } else {
+        ("addr2line", "-e")
+    };
+    let mut theirs = process::Command::new(theirs)
```
fitzgen: Can you factor all this into a `spawn_canonical_addr2line` function?

jonhoo:
@fitzgen done. ready for review.

tests/integration_test.rs
```diff
@@ -0,0 +1,133 @@
...
+        .stdin(process::Stdio::piped())
+        .stdout(process::Stdio::piped())
+        .spawn()
+        .expect("failed to spawn their addr2line");
+
+    // find a bunch of addresses
+    let names = process::Command::new("/usr/bin/nm")
+        .arg(&target)
+        .output()
+        .expect("failed to execute nm");
```
fitzgen: Can you combine the `t` filtering and every fifth address filtering with this `nm`ing into a function named `get_test_addresses`? That will really improve the readability for me here, and clearly separate gathering test data from feeding it into the `addr2line` executables.

tests/integration_test.rs
```diff
@@ -0,0 +1,133 @@
...
+    let mut theirs_out = BufReader::new(theirs.stdout.take().unwrap()).lines();
+    let mut all = 0;
+    let mut excusable = 0;
+    let mut got = 0;
+    for addr in mappings {
+        // read the oracle ouput
+        let theirs = theirs_out.next().unwrap().unwrap();
+
+        // parse it if necessary
+        let theirs = if cfg!(target_os = "macos") {
```
fitzgen: Can you factor this parsing out into its own function that returns a tuple `(Option<String>, Option<usize>)` for filename and line number (or `None` when they are `?`).

fitzgen:
Oh, also: could you squash the commits together? Thanks!

jonhoo:
@fitzgen done and done.

PR #6, opened 2016-12-04
Use a simpler but equivalent `.travis.yml` file

jonhoo:
The current `.travis.yml` file contains a number of unnecessary elements that this PR removes. This decreases overall build time, as well as getting rid of unneeded complexity.

In particular:
 - `sudo: false` is the default for all new Travis builds,
 - `osx_image: xcode7.1` is no longer valid -- the default is `xcode7.3`, and all lower versions are automatically bumped.
 - `brew update` and `brew install python` on osx: not needed, as `python2.7` is always installed on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,10 +24,9 @@ before_script:
 - if [[ -e ~/.local/bin ]]; then export PATH=~/.local/bin:$PATH; fi
 
 script:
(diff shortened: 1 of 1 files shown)


fitzgen:
Thanks! Looks great!

PR #7, opened 2016-12-04
Add support for finding functions for addresses

jonhoo:
This PR adds support for returning information about the function that contains a given address, in addition to the file and line numbers. Since supporting this means we have to store and search more DIEs, the user may choose whether or not they want function information by constructing a `Mapping` with either `with_functions` or just `new`. Current benchmarks suggest that we still do okay in terms of performance:

```console
$ addr2line
real	0m1.392s
user	0m1.370s
sys	0m0.017s
$ target/release/addr2line
real	0m1.592s
user	0m1.587s
sys	0m0.000s
$ addr2line -f
real	0m1.365s
user	0m1.340s
sys	0m0.023s
$ target/release/addr2line -f
real	0m1.661s
user	0m1.643s
sys	0m0.010s
```

It is unclear if we can keep using a `Vec` for the `units` with this setup, as it may become quite expensive to search if there are many defined functions. The `btreeranges` branch has some initial support for using a `BTreeMap` and double-ended, bounded iterators to speed up lookups (for units, not within units), though it currently requires nightly (since it needs `Bound`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,11 +4,12 @@
 //! number are associated with it.
 //!
 //! When used as a library, executables are loaded using `Mapping::new`, and users can the use
-//! `Mapping::locate` to find the corresponding file path and line number. `addr2line` avoids
-//! re-executing work where it can so that looking up multiple addresses for a single file does not
-//! require re-parsing the executable's debug symbols. The library makes heavy use of
-//! [gimli](https://github.com/gimli-rs/gimli), which provides zero-copy access to the DWARF
-//! debugging format used by most modern compilers.
+//! `Mapping::locate` to find the corresponding file path and line number. If the `Mapping` was
+//! created with `with_functions`, `Mapping::locate` may also return information about the function
+//! containing the given address. `addr2line` avoids re-executing work where it can so that looking
+//! up multiple addresses for a single file does not require re-parsing the executable's debug
+//! symbols. The library makes heavy use of [gimli](https://github.com/gimli-rs/gimli), which
+//! provides zero-copy access to the DWARF debugging format used by most modern compilers.
 //!
 //! The initial implementation of the library is heavily influenced by the original `addr2line`
 //! example in gimli, which was removed in
@@ -31,7 +32,8 @@ use object::Object;
 use owning_ref::OwningHandle;
 use fallible_iterator::FallibleIterator;
 
-use std::path::Path;
+use std::path;
+use std::borrow::Cow;
 
 pub mod error;
 pub use error::*;
@@ -63,7 +65,7 @@ struct DebugInfo<'object, Endian>
     where Endian: gimli::Endianity
 {
     debug_line: gimli::DebugLine<'object, Endian>,
-    units: Vec<Unit<'object>>,
+    units: Vec<Unit<'object, Endian>>,
 }
 
 impl Mapping {
@@ -71,7 +73,26 @@ impl Mapping {
     ///
     /// The target file will be memmap'd, and then `gimli` is used to parse out the necessary debug
     /// symbols, without copying data when possible.
-    pub fn new(file_path: &Path) -> MappingResult<Mapping> {
+    ///
+    /// The returned `Mapping` will also include debug information for functions, enabling
+    /// `Mapping::locate()` to also indicate what function a given address appears in. This comes
+    /// at some parsing and lookup cost.
+    pub fn with_functions(file_path: &path::Path) -> MappingResult<Mapping> {
+        Self::new_inner(file_path, true)
+    }
+
+    /// Construct a new `Mapping` from the debug symbols in the given executable.
+    ///
+    /// The target file will be memmap'd, and then `gimli` is used to parse out the necessary debug
+    /// symbols, without copying data when possible.
+    ///
+    /// Note that this constructor will *not* include function mapping information, and thus
+    /// `Mapping::locate()` will always return `None` for the function mapping of addresses.
+    pub fn new(file_path: &path::Path) -> MappingResult<Mapping> {
+        Self::new_inner(file_path, false)
+    }
+
+    fn new_inner(file_path: &path::Path, with_functions: bool) -> MappingResult<Mapping> {
         let file = memmap::Mmap::open_path(file_path, memmap::Protection::Read)
             .map_err(|e| MappingError::BadPath(e))?;
 
@@ -86,7 +107,7 @@ impl Mapping {
 
                         // unwrap here is currently necessary
                         // awaiting https://github.com/Kimundi/owning-ref-rs/issues/19
-                        Box::new(Self::symbolicate(file)
+                        Box::new(Self::symbolicate(file, with_functions)
                             .expect("not all necessary debug symbols are present"))
                     }),
                 })
@@ -95,21 +116,26 @@ impl Mapping {
     }
 
     /// Locate the source file and line corresponding to the given virtual memory address.
-    pub fn locate(&self, addr: u64) -> Option<(String, u64)> {
+    ///
+    /// If the `Mapping` was constructed with `with_functions`, information about the containing
+    /// function may also be returned when available.
+    pub fn locate(&self, addr: u64) -> Option<(path::PathBuf, u64, Option<Cow<str>>)> {
         self.inner.locate(addr)
     }
 
-    fn symbolicate<'a>(file: &'a object::File) -> MappingResult<EndianDebugInfo<'a>> {
+    fn symbolicate<'a>(file: &'a object::File,
+                       with_functions: bool)
+                       -> MappingResult<EndianDebugInfo<'a>> {
         if file.is_little_endian() {
-            Ok(EndianDebugInfo::LEInfo(DebugInfo::new(file)?))
+            Ok(EndianDebugInfo::LEInfo(DebugInfo::new(file, with_functions)?))
         } else {
-            Ok(EndianDebugInfo::BEInfo(DebugInfo::new(file)?))
+            Ok(EndianDebugInfo::BEInfo(DebugInfo::new(file, with_functions)?))
         }
     }
 }
 
 impl<'object> EndianDebugInfo<'object> {
-    fn locate(&self, addr: u64) -> Option<(String, u64)> {
+    fn locate(&self, addr: u64) -> Option<(path::PathBuf, u64, Option<Cow<str>>)> {
         match *self {
             EndianDebugInfo::LEInfo(ref dbg) => dbg.locate(addr),
             EndianDebugInfo::BEInfo(ref dbg) => dbg.locate(addr),
@@ -120,7 +146,9 @@ impl<'object> EndianDebugInfo<'object> {
 impl<'object, Endian> DebugInfo<'object, Endian>
     where Endian: gimli::Endianity
 {
-    fn new<'a>(file: &'a object::File) -> MappingResult<DebugInfo<'a, Endian>> {
+    fn new<'a>(file: &'a object::File,
+               with_functions: bool)
+               -> MappingResult<DebugInfo<'a, Endian>> {
         let debug_info = file.get_section(".debug_info")
             .ok_or(MappingError::MissingDebugInfo(".debug_info"))?;
         let debug_info = gimli::DebugInfo::<Endian>::new(debug_info);
@@ -139,7 +167,12 @@ impl<'object, Endian> DebugInfo<'object, Endian>
         let mut headers = debug_info.units();
         while let Some(header) = headers.next()
             .map_err(|_| MappingError::MissingDebugInfo("couldn't get DIE header"))? {
-            if let Some(unit) = Unit::parse(&debug_abbrev, &debug_ranges, &debug_str, &header) {
+            let unit = Unit::parse(&debug_abbrev,
+                                   &debug_ranges,
+                                   &debug_str,
+                                   &header,
+                                   with_functions);
+            if let Some(unit) = unit {
                 units.push(unit);
             }
         }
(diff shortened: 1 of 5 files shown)


jonhoo:
Now that we also follow inlined functions, the performance drop is somewhat higher :/

```console
$ addr2line -f
real	0m1.337s
user	0m1.310s
sys	0m0.023s
$ target/release/addr2line -f
real	0m3.387s
user	0m3.337s
sys	0m0.047s
```

This is because there are now many more units (since all the function units are also added), and they *all* have to be scanned (since there may be multiple function entries matching a given address, and we need to find *the most specific one*). The `BTreeMap` solution should help significantly with this.

I'm not entirely sure why the tests are failing. They pass consistently on my machine. When running against other binaries and comparing the output with `addr2line`, there *are* some files/lines where we output different function names, but I cannot for the life of me figure out why. Some places it looks like our output is even entirely wrong. I know the diff is pretty large, but I'd really appreciate someone else's eyes on this if someone has time to kill.

src/lib.rs
```diff
@@ -150,13 +212,68 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                    func = if dist == pdist {
+                        // we're equally good -- are we tighter?
+                        if range.end < prange.end {
+                            // yes!
+                            Some((u, range, dist))
+                        } else if range.end == prange.end {
+                            println!("found two equivalent ranges for {:x}: {} and {}",
+                                     addr,
+                                     prev.name.to_string_lossy(),
+                                     u.name.to_string_lossy());
```
philipc: Does this ever happen? Probably should use one of the logging crates for messages like this in the library code, instead of printing to stdout.

tests/integration_test.rs
```diff
@@ -80,8 +80,93 @@ fn identity_map() {
...
+#[test]
+#[cfg(not(target_os = "macos"))]
+fn with_functions() {
+    // Ignored on macOS since atos doesn't support -f
+
+    let target = env::current_exe().unwrap();
+    let mut debug = target.clone();
+
+    // On macOS, symbols are kept in a separate .dSYM file
+    if cfg!(target_os = "macos") {
```
Mark-Simulacrum: This will do nothing, since this function isn't compiled for `target_os = "macos"`.

This seems to happen a few more times in this function, too.

tests/integration_test.rs
```diff
@@ -80,8 +80,93 @@ fn identity_map() {
...
+                let f = f.file_name().unwrap().to_string_lossy().into_owned();
+                file = f;
+            }
+
+            assert_eq!(oracle.0, Some(&*file));
+            assert_eq!(oracle.1, Some(lineno));
+
+            if let Some(func) = func {
+                // We even tried to guess the function!
+                assert_eq!(function, func);
```
Mark-Simulacrum: As far as I can tell, this is the failing test line, which seems to be failing because of potentially different demangling:
`_ZN100_$LT$core..slice..Iter$LT$$u27$a$C$$u20$T$GT$$u20$as$u20$core..iter..traits..DoubleEndedIterator$GT$9next_back17h965518d3b70e08caE` looks quite similar to`_ZN4core5slice8{{impl}}13next_back<u8>E`.

But it's probably not just demangling; though similar enough that it's suspicious.

src/lib.rs
```diff
@@ -192,59 +312,136 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
-// TODO: most of this should be moved to the main library.
-struct Unit<'input> {
-    address_size: u8,
-    ranges: Vec<gimli::Range>,
-    line_offset: gimli::DebugLineOffset,
-    comp_dir: Option<&'input std::ffi::CStr>,
-    comp_name: Option<&'input std::ffi::CStr>,
+enum Target<'input> {
+    Line(LineUnit<'input>),
+    Function(FunctionUnit<'input>),
```
philipc: Is there any reason these need to be in the same enum, instead of having separate `Vec<LineUnit>` and `Vec<FunctionUnit>`? There's probably a small memory saving from having them separate.

src/lib.rs
```diff
@@ -294,9 +495,34 @@ impl<'input> Unit<'input> {
...
 
+    fn ranges(&self) -> &[gimli::Range] {
+        match *self {
+            Target::Function(ref u) => &u.ranges[..],
+            Target::Line(ref u) => &u.ranges[..],
+        }
+    }
+}
+
+struct FunctionUnit<'input> {
```
philipc: Not sure it makes sense to include `Unit` in the names for these. The original `struct Unit` was the info for a compilation unit. Maybe `FunctionInfo/LineInfo` or something like that?

tests/integration_test.rs
jonhoo: True. My hope is that eventually we'll find a way to run that test under macOS too, and I kept it in there assume that we'd need to perform the same hacks when that was the case. But I guess you're right that we should rather just take them out for now.

tests/integration_test.rs
jonhoo: Yes, that one is certainly similar (specifically, one looks like a specialization of the other). I believe this is because we don't correctly detect inlining in all cases (or we detect it differently from `addr2line` does). I have also observed other cases where we actually produce entirely *different* function names, and that's more disturbing.

src/lib.rs
jonhoo: It actually does happen quite frequently. My guess is that it happens due to recursive function inlining, where the intermediate functions just forward to another function. I agree it is strange though.

The `println!`s are only there for debugging while I develop (PR is still `[WIP]` after all) -- they will be removed before merging.

src/lib.rs
jonhoo: I guess they could be separate, though this provides a neat way to share functionality between them (like looking up ranges). I guess we could provide something similar with a wrapper type though. Is there a particular reason you'd like to see them separated?

src/lib.rs
jonhoo: Agreed. Will make that change on my next pass.

jonhoo:
@philipc 

> The integration test works on my machine. For the failing Travis symbol, it seems like the Travis addr2line is getting the function name from the symbol table instead of the DWARF debug info (the name that the Travis addr2line gives is the same one that nm would give.)

Oh, curious. Travis is running with an older version of `binutils`, so that could explain it. I don't know if there's a way to use a newer `addr2line` (which is presumably what we want to compare against) on their infrastructure? I guess we could move to `dist: trusty`...

> For performance, you could consider structuring things as a tree with three levels: compilation unit, subprogram, and inlined subroutine. So you would first search for the correct compilation unit, and then you only need to search through the subprograms for that compilation unit, and again then only search the inlined subroutines for that subprogram.

Hmm.. Are we actually guaranteed that an address will only match inside a single compilation unit? And that the corresponding subprogram range is contained within that same compilation unit? If that is the case, that might help a little. Unfortunately, most of the programs I've been testing with so far are dominated by the size of a particular compilation unit (usually the main program), and so this would only reduce the search space by a small fraction. Probably worth exploring anyway. This change is orthogonal to the `BTreeMap` change I want to make, which will make searching for units `O(log n)`, as well as allow early termination of the search.

src/lib.rs
philipc: The `Line` variant is larger than the `Function` variant (88 bytes vs 56 bytes), but there'll be many more functions than compilation units, so there'll be some memory saving. Also, the usage in `locate()` seems to treat them separately, and the flow control is a bit different (you want the most specific function, but the first line match is ok), so I wasn't sure there was much benefit to having them in the same enum. The only common bit I can see is parsing the range attributes, but that could be moved to a helper function that they both call. Not something I feel strongly about though.

philipc:
> Are we actually guaranteed that an address will only match inside a single compilation unit? And that the corresponding subprogram range is contained within that same compilation unit?

Well, we can't guarantee that all DWARF is perfectly formed, but I think the intent is that these should hold.

jonhoo:
Following some very basic print debugging, it does seem like that's the case for the binaries I have available. I'll make the change.

To give an idea of the skew in compilation unit size I alluded to, here is the number of units with `n` subprograms in a `--release` compile of our `addr2line` binary:
```
n      # of compilation units
0     1
1     1
2     6
3     3
4     1
5     1
6     4
8     1
9     3
10    1
11    2
12    1
15    1
25    1
41    1
43    1
45    1
47    1
70    1
113   1
152   1
775   1
1595  1
5138  1
```

jonhoo:
Using the two-level search you proposed, we're now roughly back where we should be:

```console
$ addr2line -f
real	0m1.309s
user	0m1.280s
sys	0m0.027s
$ target/release/addr2line -f
real	0m1.674s
user	0m1.653s
sys	0m0.017s
```

I may take a stab at three-level searching tomorrow if I get the time.

jonhoo:
@philipc actually, the more I think about it, the more I believe that whatever is causing us to hit the "found two equivalent ranges" case is what causes the few remaining discrepancies I am seeing. Will have to dig further into that at some point.

jonhoo:
I think this might stem from the way we resolve `DW_AT_abstract_origin`. The spec says they are "references", but doesn't say what they are relative to. Furthermore, `func_names` map is initialized with `entry.offset().0` as the key, but I just realized that that offset is relative to the start of the current compilation unit, not global. Thus you may end up with collisions, which would yield wrong results (as we've been seeing) when looking up in it using `abstract_origin`.

We extract the abstract origin using
```rust
if let Some(gimli::AttributeValue::UnitRef(origin)) = entry.attr_value(gimli::DW_AT_abstract_origin)
```
but it's not clear to me what kind of magic that does to the reference (does it make it unit-local? global?). Nor how we can compute the corresponding offset for each `subprogream` when we iterate over them so that we can correctly match the two up...

philipc:
The `UnitOffset` is always relative to the unit containing the offset. `attr_value` passes through the raw value, there's no magic in what it does. Using those offsets for a global `func_names` isn't going to work. You only need to the `func_names` map while parsing, so I would move its creation into `Unit::parse`.

Another option would be to stop using the `func_names` map, and whenever you need the name for a `DW_AT_abstract_origin`, create a new `EntriesCursor` to read it. You'll end up parsing the same abstract origin a few times, but on the other hand you won't have to traverse the entire DIE tree twice, so it may be similar performance, not sure.

philipc:
> actually, the more I think about it, the more I believe that whatever is causing us to hit the "found two equivalent ranges" case is what causes the few remaining discrepancies I am seeing.

I haven't checked, but one possible reason for this is an inline function that contains another inline function and nothing else. eg `fn foo() { bar() }`.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                        current = Some(row.clone());
+                    }
+                    continue;
+                }
+                break;
+            }
+
+            // The row we just last iterated to is *after* the address, to the previous row we
+            // saw (stored in current) is the one we want. There better be one...
+            let row = current.unwrap();
```
fitzgen: Despite the comment, it isn't clear to me that this `unwrap` won't panic: what if there aren't any rows in the line number program?

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
                     }
-                    break;
                 }
-                if row.end_sequence() {
-                    current = None;
+                path.push(&*directory);
+            }
+            path.push(&*file.path_name().to_string_lossy());
+
+            let line = row.line().unwrap();
```
fitzgen: Why `unwrap` here? We should be able to handle an unknown line number and print `??`, right?

I guess we did an `unwrap` in the old code too. It's pretty clear that it was an example program, and not as polished as `gimli` proper :P

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                                     addr,
+                                     prev.name.to_string_lossy(),
+                                     p.name.to_string_lossy());
+                            Some((p, range, dist))
+                        } else {
+                            // no
+                            Some((prev, prange, pdist))
+                        }
+                    } else if dist < pdist {
+                        // we're just better
```
fitzgen: :-D

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                        Some((prev, prange, pdist))
+                    };
                 } else {
-                    current = Some(row.clone());
+                    func = Some((p, range, dist));
                 }
+
+                // Unfortunately, we're not done even if we've already found loc, because there
+                // *may* be another subprogram (specifically, one that has been inlined) that
+                // matches us better. We need to keep going until we find the best one.
```
fitzgen: It would be nice to (optionally) return the whole set of virtual frames (one physical frame and all the frames inlined into it) instead of only the innermost inlined frame.

A profiler that is symbolicating a stack of physical frames would certainly like to have this information...

For another PR, either way.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
-        } else if let Some(range) = Self::parse_contiguous_range(entry) {
-            vec![range]
-        } else {
-            return None;
-        };
+        let mut unit = {
+            // Scoped so that we can continue using entries for the loop below
+            let first = entries.next_dfs();
+            let (_, entry) = first.expect("Should parse first entry OK")
+                .expect("And first entry should exist!");
```
fitzgen: Now that this is a library, rather than an example program, we probably shouldn't panic on malformed DWARF, so we should probably early return `None` here rather than `expect("...")`ing that the input data is valid.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+            assert_eq!(entry.tag(), gimli::DW_TAG_compile_unit);
```
fitzgen: And ditto here about not panicking on malformed input.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+            if ranges.is_empty() {
+                continue;
+            }
+
+            if cfg!(debug_assertions) {
+                // All programs should be fully contained within their compilation unit
+                for r in &ranges {
+                    assert!(unit.ranges
+                        .iter()
+                        .any(|range| range.begin <= r.begin && range.end >= r.end));
```
fitzgen: Ditto.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                unit.programs.push(Program {
+                    ranges: ranges,
+                    inlined: entry.tag() == gimli::DW_TAG_inlined_subroutine,
+                    name: name,
+                });
+                continue;
+            }
+
+            // This really shouldn't happen
+            panic!("subprogram {:x} has neither (linkage_,)name, nor abstract origin",
```
fitzgen: Ditto about panicking.

src/lib.rs
```diff
@@ -150,110 +213,260 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                    });
+                    continue;
+                } else {
+                    panic!("failed to resolve subprogram origin {:x}", entry.offset().0);
+                }
+            }
+
+            // Linked name is not available, and we don't have a reference to another function.
+            // Fall back to just plain old name, if that's available.
+            if let Some(name) = entry.attr(gimli::DW_AT_name)
```
fitzgen: All this naming stuff is getting pretty involved -- would be nice to factor it out into smaller functions.

src/lib.rs
```diff
@@ -135,11 +164,45 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+        let mut func_names = if with_functions {
+            Some(HashMap::new())
+        } else {
+            None
+        };
+        if let Some(ref mut func_names) = func_names {
+            let mut headers = debug_info.units();
+            while let Some(header) = headers.next()
+                .map_err(|_| MappingError::MissingDebugInfo("couldn't get DIE header"))? {
+                let abbrev = header.abbreviations(debug_abbrev).expect("Fail");
```
jonhoo: This is a left-over from the original `addr2line` code as it was in gimli ;)
I'm planning on doing a subsequent PR that cleans up error handling throughout the code, but will modify this one while I'm here.

src/lib.rs
jonhoo: Yeah, this is already on my to-do list. But yes, in another PR.

jonhoo:
@philipc Oh, I didn't know about `entries_at_offset`. I'll move the code to that straight-away and get rid of `func_names`. I also think you're right about the found equivalent cases, though that still doesn't explain why we're getting incorrect results. The function name resolution for abstract origins might though!

@fitzgen As I also noted inline, I'd prefer to address all the error handling holistically in a separate PR, and have this PR keep all the `expect`s and `unwrap`s from the original code. Would that be ok?

src/lib.rs
jonhoo: Done.

jonhoo:
I believe we're just about there with b9f6701aad6c613e0587bda59f8f46efc82a069f. Some notes in the commit message. In particular, the use of `entries_at_offset` necessitates moving to a git version of gimli, since 0.9.0 doesn't have it. Once a new gimli release is cut, we can move back to non-git.

jonhoo:
Gah, spoke too soon again. I now observe *no* discrepancies between us and `addr2line` on a relatively large binary I have locally (that is, we either give no answer, or the same answer), but for some reason Travis still sees this `{impl}` specialization mismatch... Suggestions welcome!

jonhoo:
Okay, I moved to the newer Travis environment and everything passes just fine. I suspect what was happening was that the old Travis setup had an outdated version of `addr2line`, and we actually resolved *better* than it did. I consider this ready to merge.

jonhoo:
The macOS results probably won't come in for a while given [Travis' current macOS backlog](https://www.traviscistatus.com/), but as we don't run the tests there, I don't expect any problems.

philipc:
> I suspect what was happening was that the old Travis setup had an outdated version of addr2line, and we actually resolved better than it did. I consider this ready to merge.

It may actually be a bug in rustc if the DWARF subprogram linkage name doesn't match the ELF symbol table. I don't know much about mangling, but I didn't think `{}<>` were valid characters in the linkage name.

jonhoo:
Given that it matches fine with a newer version of `addr2line`, I'm hesitant if that's the right conclusion to draw, but maybe.

philipc:
I meant that as more of an aside. I think the behaviour of the newer addr2line is better, but I also think that the DWARF linkage name is a bit strange.

jonhoo:
Ah, I see. Yes, I agree. Related to that again, we may want to add support for symbol de-mangling at some point, though I'm unsure what the best way to do so is. Ideally that should be in an entirely separate crate...

jonhoo:
Seems like there is one for [Rust](https://github.com/alexcrichton/rustc-demangle), but not one that covers other languages as well..

fitzgen:
Thanks for following through with this, @jonhoo !

Regarding demangling: would prefer leveraging other crates (and contributing as needed).

PR #9, opened 2016-12-05
Provide better error-handling code within the library

jonhoo:
Following feedback in #7, this PR gets rid of all calls to `unwrap` and `expect` that may panic, and instead optimistically continues. All such cases have been annotated with

```rust
/// DWARF is invalid -- <reason>
```

I will rebase this onto master once #7 and #8 have been merged.


jonhoo:
Should we add in `debug_assert!` statements for each of the cases where we currently encounter invalid data? Might help in diagnosing errors further down the line?

jonhoo:
Alternatively, we could make `locate()` return a `Result`, though `Result<Option<T>>` isn't a great type to expose through the API. Maybe the inner `locate()` methods could return that, and `Mapping::locate()` could flatten it into just an `Option`? And then we could add a `debug_assert` there?

jonhoo:
8b07f95c623013e4832a6c1ee83d963120643dc7 uses [error-chain](https://github.com/brson/error-chain) to implement deep error handling, propagating error chains all the way up to the `pub` methods on `Mapping`. `Mapping::locate()` currently simply drops the `Err` if there is one, but in the future this would be where we'd add in logging or debug assertions.

src/lib.rs
```diff
@@ -213,12 +213,26 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+            if current.is_none() {
+                return None;
+            }
             let row = current.unwrap();
             let header = rows.header();
 
+            let file = row.file(header);
+            if file.is_none() {
+                // DWARF is invalid -- DebugLine refers to file that does not exist!
+                return None;
```
fitzgen: I would generally prefer to define a `enum Error { ... }` and define variants for each of these `return None` cases, eg `Error::ReferenceToUnknownFile`.

src/lib.rs
```diff
@@ -243,6 +257,7 @@ impl<'object, Endian> DebugInfo<'object, Endian>
                 }
 
                 // This program covers the given address -- calculate how well it matches
+                // NOTE: the unwrap is okay, because p.contains_address returned true above.
```
fitzgen: Rather than commenting `unwrap()` calls, I prefer to pass what would have been a comment into an `expect("...")` call. That way, if a panic does happen, we have more of a clue what is going on.

src/lib.rs
```diff
@@ -454,14 +491,26 @@ impl<'input, Endian> Unit<'input, Endian>
             Some(gimli::AttributeValue::Addr(addr)) => addr,
             _ => 0,
         };
-        let ranges = debug_ranges.ranges(offset, address_size, base_address)
-            .expect("Range offset should be valid");
-        Some(ranges.collect().expect("Should parse ranges"))
+
+        if let Ok(ranges) = debug_ranges.ranges(offset, address_size, base_address) {
+            if let Ok(ranges) = ranges.collect() {
+                return Some(ranges);
```
fitzgen: This whole nested `if let` could be rewritten a little more idiomatically like this:

```rust
debug_ranges.ranges(offset, address_size, base_address)
    .and_then(|ranges| ranges.collect())
    .ok()
```

jonhoo:
@fitzgen ready for review

src/lib.rs
```diff
@@ -27,16 +27,92 @@ extern crate memmap;
...
+use std::fmt;
 use std::path;
+use std::error;
 use std::borrow::Cow;
 
-pub mod error;
-pub use error::*;
+/// An error occured while traversing the debug symbols in the provided executable.
+#[derive(Debug)]
+pub enum DebugInfoError {
```
fitzgen: Any reason why this is a separate type from the `error` module's `Error` rather than just having one common error type for the whole library?

src/lib.rs
```diff
@@ -118,13 +194,16 @@ impl Mapping {
     ///
     /// If the `Mapping` was constructed with `with_functions`, information about the containing
     /// function may also be returned when available.
-    pub fn locate(&self, addr: u64) -> Option<(path::PathBuf, u64, Option<Cow<str>>)> {
-        self.inner.locate(addr)
+    pub fn locate(&self, addr: u64) -> Option<(path::PathBuf, Option<u64>, Option<Cow<str>>)> {
+        self.inner.locate(addr).unwrap_or_else(|_| {
+            // TODO: we probably want to log this, or assert failure when testing?
```
fitzgen: Why not return the error and propagate it up the stack where the caller can decide what is appropriate?

jonhoo: I discussed this briefly in the comments on the PR above, but it is basically to expose a simpler API to the user. In general, they probably want errors to simply return `None` as the location, rather than having to do error matching themselves. That said, forcing the user to do `unwrap_or(None)` isn't all that heavy a burden. I guess I can change it.

jonhoo: Addressed in 277d72cf3644856694cbde03b92d4ba8da169853

src/lib.rs
fitzgen: Ok, that makes sense, thanks.

src/lib.rs
philipc: Note that `ok()` is shorthand for `unwrap_or(None)`.

jonhoo: Ah, yes, good catch.

PR #12, opened 2016-12-06
Write up a better README

jonhoo:
Includes more information about the library, how to quickly add it to your project, and a detailed usage example for both the library and the binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,129 @@
-# `addr2line`
+# addr2line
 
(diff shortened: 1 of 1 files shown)


jonhoo:
Note that this probably shouldn't be merged until #7 has also been merged, as the `README` example discusses functions and relies on a three-tuple being returned from `Mapping::locate`.

Issue #2, opened 2016-12-04
addr2line is slower than the binutils version


philipc:
From https://github.com/gimli-rs/gimli/issues/53.

One probable optimization will be to cache the parsed `.debug_line` rows.

jonhoo:
Another I'm looking into right now is to use a `BTreeMap` indexed by base address to keep the `Unit`s in. By then looking into the range `(-, addr]` and stepping backwards through the iterator, we should be much more efficient about doing the lookups.

jonhoo:
One thing that would speed things up significantly is if we had a way to binary search the line rows within a particular `Unit`. Currently, we scan left to right and stop when the next entry is larger than the address we're looking for, but if we could efficiently skip backwards and forwards (like in an array), this would be much faster. @fitzgen, is there a way to make the `StateMachine` that's returned by `gimli::LineNumberProgramHeader::rows()` allow efficient seeking, not just one-at-a-time processing?

fitzgen:
> @fitzgen, is there a way to make the StateMachine that's returned by gimli::LineNumberProgramHeader::rows() allow efficient seeking, not just one-at-a-time processing?

I don't think so. The address, line number, source file, etc matrix is encoded and compressed into a bytecode program, and `StateMachine` is interpreting that program to yield one row of the matrix at a time. Each row of the matrix requires information from the previous one, since (per spec) all the bytecodes mutate the current row being reconstructed in place, and often deal with relative numbers and deltas rather than absolute numbers, since they can be used to encode the matrix more compactly.

fitzgen:
Also, FWIW, it isn't clear that we are still really slower than the canonical binutils `addr2line`. We would certainly benefit from more benchmarking.

jonhoo:
And from profiling, most of the time seems to be spent in iterating over `rows()`. It's unfortunate we can't skip through it.. I don't really want to extract them all at start-up time, as that would waste significant memory, plus slow down applications that *don't* do a lot of lookups. Maybe a simple addr->file cache is sufficient.

fitzgen:
Thanks for doing additional profiling!

jonhoo:
[...]

PR #14, opened 2016-12-09
Implement lazy binary search for line rows

jonhoo:
This PR implements one of the optimizations discussed in #2. In particular, it lazily instantiates a list of `gimli::StateMachine` every `n` (currently 100) `LineNumberRow` as it walks the line rows for a `Unit`. For each cached row, it also tracks the address the state machine encountered at that point, allowing us to binary search over this "skiplist" when we're given a new address, rather than always doing a linear scan from the start of the unit. The speed-up is quite significant: ~100x for file/line lookups, and ~10x for file/line/function lookups.

```console
$ objdump --dwarf=info target/release/vote | grep -E 'subprogram|subroutine' | wc -l
92572
$ objdump -C -x --disassemble -l target/release/vote | grep -P '0[048]:' | awk '{print $1}' | sed 's/:$//' > addresses.txt
$ wc -l addresses.txt 
3140 addresses.txt
$ cat bench.sh
#!/bin/sh
for f in "" "-f"; do
	for bin in addr2line target/release/addr2line-master target/release/addr2line; do
		echo "$bin $f";
		time "$bin" -a -e target/release/vote $f < addresses.txt
	done
done
$ ./bench.sh # I cleaned up the output a little
addr2line
real	0m1.168s
user	0m1.137s
sys	0m0.030s
target/release/addr2line-master
real	0m1.211s
user	0m1.203s
sys	0m0.007s
target/release/addr2line
real	0m0.037s
user	0m0.013s
sys	0m0.023s

addr2line -f
real	0m1.170s
user	0m1.127s
sys	0m0.040s
target/release/addr2line-master -f
real	0m1.415s
user	0m1.393s
sys	0m0.020s
target/release/addr2line -f
real	0m0.209s
user	0m0.187s
sys	0m0.020s
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,7 @@ use owning_ref::OwningHandle;
 use fallible_iterator::FallibleIterator;
 
 use std::fmt;
+use std::sync;
 use std::path;
 use std::error;
 use std::borrow::Cow;
@@ -142,6 +143,7 @@ struct DebugInfo<'object, Endian>
 {
     debug_line: gimli::DebugLine<'object, Endian>,
     units: Vec<Unit<'object, Endian>>,
+    cache_every: usize,
 }
 
 impl Mapping {
@@ -255,6 +257,7 @@ impl<'object, Endian> DebugInfo<'object, Endian>
 
         Ok(DebugInfo {
             debug_line: debug_line,
+            cache_every: 100,
             units: units,
         })
     }
@@ -268,21 +271,93 @@ impl<'object, Endian> DebugInfo<'object, Endian>
                 continue;
             }
 
+            let mut rowi = 0;
+            let mut current = None;
+
             // Okay, this is the right unit. Check our DebugLine rows.
-            let mut rows = unit.line_rows(&self.debug_line)
-                .chain_err(|| "cannot get line rows for unit")?;
+            let mut rows = {
+                // First, check our binary search list
+                if let Ok(skiplist) = unit.skiplist.read() {
+                    match skiplist.binary_search_by_key(&addr, |&(raddr, _, _)| raddr) {
+                        Ok(i) => {
+                            // we have a state machine for this address!
+                            current = Some(skiplist[i].2);
+                            rowi = (i + 1) * self.cache_every + 1;
+                            skiplist[i].1.clone()
+                        }
+                        Err(i) => {
+                            // i is the entry in the skiplist where addr *would* appear. Thus, we
+                            // don't have a state machine for *this* address, but we do know that:
+                            //
+                            //  - if i == 0, we have no state machine, and must from scratch
+                            //  - if i == skiplist.len(), we scan from the last state machine
+                            //  - otherwise, the state machine from i-1 eventually encounters addr
+                            if i == 0 {
+                                unit.line_rows(&self.debug_line)
+                                    .chain_err(|| "cannot get line rows for unit")?
+                            } else {
+                                current = Some(skiplist[i - 1].2);
+                                // NOTE
+                                // we need to use i * self.cache_every here, not (i+1) as above.
+                                // this is because i is the i *after* the one we're chooseing to
+                                // start from (skiplist[i-1] above).
+                                rowi = i * self.cache_every + 1;
+                                skiplist[i - 1].1.clone()
+                            }
+                        }
+                    }
+                } else {
+                    // the skiplist was poisoned -- fall back to linear scan
+                    unit.line_rows(&self.debug_line)
+                        .chain_err(|| "cannot get line rows for unit")?
+                }
+            };
 
             // Now, find the last row before a row with a higher address than the one we seek.
-            let mut current = None;
-            while let Ok(Some((_, row))) = rows.next_row() {
-                if row.address() <= addr {
-                    if row.end_sequence() {
-                        current = None;
-                    } else {
-                        // Might be the right row, but we won't know until we see the next one.
-                        // The .clone is needed so we can keep iterating
-                        current = Some(*row);
+            let mut praddr = 0;
+            let mut skipseq = false;
+            while let Ok(Some((_, &row))) = rows.next_row() {
+                if row.end_sequence() {
+                    current = None;
+                    skipseq = false;
+                    continue;
+                }
+
+                if skipseq {
+                    continue;
+                }
+
+                let raddr = row.address();
+                if raddr < praddr {
+                    // NOTE:
+                    // We currently skip these sequences, but we *should* of course handle them
+                    // correctly. It's unclear how the interplay between this and the skiplist
+                    // should work.
+                    // unimplemented!();
+                    skipseq = true;
+                    continue;
+                }
+                praddr = raddr;
+
+                if raddr <= addr {
+                    // Might be the right row, but we won't know until we see the next one.
+                    // The .clone is needed so we can keep iterating
+                    current = Some(row);
+
+                    // Add every self.cache_every'th non-empty row to the skiplist
+                    if rowi != 0 && rowi % self.cache_every == 0 {
+                        if let Ok(mut skiplist) = unit.skiplist.write() {
+                            let i = rowi / self.cache_every - 1;
+                            if i >= skiplist.len() {
+                                debug_assert!(i == skiplist.len(),
+                                              "we somehow didn't cache a StateMachine for a \
+                                               previous iteration step!");
+                                // cache this StateMachine
+                                skiplist.push((row.address(), rows.clone(), row));
+                            }
+                        }
                     }
+                    rowi += 1;
                     continue;
                 }
                 break;
@@ -386,7 +461,11 @@ impl<'object, Endian> DebugInfo<'object, Endian>
 
 // TODO: most of this should be moved to the main library.
 use std::marker::PhantomData;
-struct Unit<'input, Endian> {
+struct Unit<'input, Endian>
+    where Endian: gimli::Endianity
+{
+    skiplist: sync::RwLock<Vec<(u64, gimli::StateMachine<'input, Endian>, gimli::LineNumberRow)>>,
+
     address_size: u8,
     ranges: Vec<gimli::Range>,
     line_offset: gimli::DebugLineOffset,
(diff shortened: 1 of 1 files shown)


jonhoo:
The 100 is currently hard-coded, but as @fitzgen proposed, it should probably be `sqrt(n)` for each unit. Is there an efficient way for us to figure out the `n` for a unit without scanning it start-to-finish? Alternatively, if we're willing to scan it, we could eagerly create the skiplist...

jonhoo:
Oh, curious. I can reproduce the failure locally with nightly — will look into it once I get the chance.

fitzgen:
Have you tried different values of N to see how performance is affected? N=1, in particular would be interesting.

jonhoo:
No, only N=100, but I'd want a performance/N chart of this before merging, so it's on the roadmap.

fitzgen:
> Is there an efficient way for us to figure out the n for a unit without scanning it start-to-finish?

No, it requires parsing and evaluation the bytecodes. This was why I suggested looking at the avg bytes-per-row in the wild and then using that as a way to estimate N in `addr2line`.

jonhoo:
So there's a way for us to find the number of bytes of row data for a given unit?

fitzgen:
@jonhoo I was imagining instrumenting a local build of `gimli` to 

1. record the number of bytes in the line number program file
2. record the number of rows in the line number program file

And then use that info with a few different Real World(tm) binaries and look at (1) / (2) to come up with a good rule of thumb / estimate for bytes-per-row.

jonhoo:
I understand. But for that ratio to be useful (that is, when an end-user runs `addr2line`), we'd also need to be able to get the number of bytes of rows for a given unit in that case, so that we can divide it by the ratio we've set to get the number of rows.

philipc:
It's the length of `LineNumberProgramHeader::program_buf`, but that isn't public.

fitzgen:
> but that isn't public.

Which we can always change, of course ;)

jonhoo:
Some results for my application:

Runtime of different variations compared to binutils `addr2line` and `master`:
time

Memory usage:
memory

Runtime as `n` is varied (basically the same information as is in the first plot):
vary-n

jonhoo:
Yeah, exposing `LineNumberProgramHeader::program_buf` should probably be sufficient.

fitzgen:
Wow! This is great!

Looks like N=100 was a pretty good guess!

jonhoo:
This is only on my one binary though, so I don't know if it generalizes. I'll see if I can dig up some other decently-sized binaries to try it on.

jonhoo:
Different binary (C++ this time):
```console
$ objdump --dwarf=info at | grep -E 'subprogram|subroutine' | wc -l
36237
$ objdump -C -x --disassemble -l at | grep -P '[048]:' | awk '{print $1}' | sed 's/:$//' > addresses.txt
$ wc -l addresses.txt 
9708 addresses.txt
```
memory
time
vary-n

jonhoo:
I guess as a starting point we could make `LineNumberProgramHeader::program_buf` public, and then see what ratio gives us `n=100` for my two programs. I could then code that in, and it'd be easier to test performance + memory usage on other programs. I could also add a flag to the binary to let the user choose whether or not to do this lazy caching, which would allow easy testing against a baseline.

philipc:
How are you measuring memory usage?

jonhoo:
```console
$ /usr/bin/time -f '%e\t%M'
```

philipc:
The test is failing; do you have any ideas why?

One problem: there is no guarantee that the line number information is sequential by address. I'm not sure that this is fixable when using a skiplist.

jonhoo:
No, but I haven't had a chance to dig into it yet. I can reproduce the failure locally with nightly though. If we're not guaranteed that the line number if sequential by address, then the current implementation is also wrong (it walks the rows until it finds the first with an address higher than the target)

philipc:
> If we're not guaranteed that the line number if sequential by address, then the current implementation is also wrong (it walks the rows until it finds the first with an address higher than the target)

Yeah I think it's wrong too, but it's fixable by not stopping early. Note that the fact that it succeeds means the executables we've been testing on are sequential, so maybe we can optimize for that case still. If I dump the line information with objdump, then I see some sequences that start at address 0, such as:
```
/buildslave/rust-buildbot/slave/nightly-dist-rustc-linux/build/obj/../src/libcore/slice.rs:
slice.rs                                    1974                   0
slice.rs                                    1974                0x10
slice.rs                                    1975                0x20
slice.rs                                    1976                0x34
```
I think these are invalid sequences, but if the binary search hits one of them, it'll get confused. They don't affect the current implementation because they are lower than the address we are searching for.

So maybe this is fixable with 2 changes:
- ignore sequences that start at 0
- stop using binary search if a sequence starts at a lower address than the previous sequence (or at least put in an assert for this case)

Note that DWARF does guarantee that:
> Within a sequence, addresses (operation pointers) may only increase.

jonhoo:
@philipc It actually turns out that the bug was that I also cached rows where `row.end_sequence() == true`, where obviously `row.address() == 0`. That is fixed in 853632f. 24ddf27 rewrites the code somewhat so the logic is more evident. It also starts tracking the previous address to detect if we ever jump backwards in addresses. The call to `unimplemented!()` is currently commented out since it turns out this *does* happen (which might be why we're missing a bunch of debug symbols compared to `addr2line`). 2147e1d changes the behavior such that we skip a unit once we detect this situation for a particular address, since we know we won't handle it correctly.

jonhoo:
Oh, I'll deal with the new gimli interface in a separate PR

jonhoo:
Needs to be rebased once #15 is merged.

jonhoo:
Note that this early-breaking does mean that we resolve fewer addresses than before:
```console
$ cargo t -- --nocapture | grep resolved
resolved 703/839 addresses (136 file-only misses)
resolved 703/839 addresses (136 file-only misses, 703 function hits)
$ git checkout bsearch-lines
$ cargo t -- --nocapture | grep resolved
resolved 407/914 addresses (152 file-only misses)
resolved 407/914 addresses (152 file-only misses, 407 function hits)
```

I think the way to resolve this is a separate PR (which I'm already working on) that will simply skip *sequences* that jump backwards, but continue with the unit overall.

jonhoo:
Never mind, fixed in 71322c6:
```console
$ cargo t -- --nocapture | grep resolved
resolved 762/914 addresses (152 file-only misses)
resolved 762/914 addresses (152 file-only misses, 762 function hits)
```

fitzgen:
@jonhoo is this ready for a proper review now?

jonhoo:
Rebased, squashed, and ready for review.
r? @fitzgen @philipc

src/lib.rs
```diff
@@ -268,21 +271,95 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                    continue;
+                }
+
+                if skipseq {
+                    continue;
+                }
+
+                let raddr = row.address();
+                if raddr == 0 {
+                    // NOTE: It is unclear if these rows are valid.
```
philipc: I think we should set `skipseq = true` for this case too.

src/lib.rs
```diff
@@ -268,21 +271,95 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
-            let mut rows = unit.line_rows(&self.debug_line)
-                .chain_err(|| "cannot get line rows for unit")?;
+            let mut rows = {
+                // First, check our binary search list
+                if let Ok(skiplist) = unit.skiplist.read() {
+                    match skiplist.binary_search_by_key(&addr, |&(raddr, _, _)| raddr) {
+                        Ok(i) => {
+                            // we have a state machine for this address!
+                            current = Some(skiplist[i].2);
+                            rowi = (i + 1) * self.cache_every;
```
philipc: The skiplist holds the state machine from after reading a row that has an index that is a multiple of `cache_every`, which means the next row that state machine gives will be one more than that, so this should be `rowi = (i + 1) * self.cache_every + 1;`.

src/lib.rs
```diff
@@ -268,21 +271,95 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                            //
+                            //  - if i == 0, we have no state machine, and must from scratch
+                            //  - if i == skiplist.len(), we scan from the last state machine
+                            //  - otherwise, the state machine from i-1 eventually encounters addr
+                            if i == 0 {
+                                unit.line_rows(&self.debug_line)
+                                    .chain_err(|| "cannot get line rows for unit")?
+                            } else {
+                                current = Some(skiplist[i - 1].2);
+                                rowi = (i + 1) * self.cache_every;
```
philipc: Ditto.

src/lib.rs
```diff
@@ -268,21 +271,95 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                }
+                praddr = raddr;
+
+                if raddr <= addr {
+                    // Might be the right row, but we won't know until we see the next one.
+                    // The .clone is needed so we can keep iterating
+                    current = Some(row);
+
+                    // We only cache non-end rows, because only there do we know that
+                    // .address() isn't 0.
```
philipc: I don't understand this comment. I thought we handled end sequence and zero address already. Is it leftover from a prior version?

src/lib.rs
jonhoo: Hmm, the question is if a single zero entry should invalidate the entire sequence... I don't know why we should require this? Without `skipseq = true` here, we'll simply move to the next entry, and if that is still lower, then we'll skip the sequence...

philipc: Why skip the zero either then?

src/lib.rs
jonhoo: You're right. Fixed.

src/lib.rs
jonhoo: True. I guess the right thing to do is to simply get rid of this if, and instead let the next `raddr < praddr` branch take care of it.

jonhoo:
Great! Squashed and ready to merge. Fixes #2?

philipc:
Yep fixes #2, feel free to merge when checks have passed.

jonhoo:
@philipc I don't have write access to the repo it seems, so I can't merge..?

philipc:
I merged, but you should have write access too now.

fitzgen:
\o/

jonhoo:
Note that this is still using a fixed `n=100`, not `LineNumberProgramHeader::program_buf`.

PR #15, opened 2016-12-11
Handle gimli exposing errors from attr()

jonhoo:
Fixes code to work with changes introduced in gimli-rs/gimli#165.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -433,12 +433,20 @@ impl<'input, Endian> Unit<'input, Endian>
 
             // Extract source file and line information about the compilation unit
             let line_offset = match entry.attr_value(gimli::DW_AT_stmt_list) {
-                Some(gimli::AttributeValue::DebugLineRef(offset)) => offset,
+                Ok(Some(gimli::AttributeValue::DebugLineRef(offset))) => offset,
+                Err(e) => {
+                    return Err(Error::from(ErrorKind::Gimli(e)))
+                        .chain_err(|| "invalid compilation unit statement list")
+                }
                 _ => return Ok(None),
             };
             let comp_dir = entry.attr(gimli::DW_AT_comp_dir)
+                .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+                .chain_err(|| "invalid compilation unit directory")?
                 .and_then(|attr| attr.string_value(debug_str));
             let comp_name = entry.attr(gimli::DW_AT_name)
+                .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+                .chain_err(|| "invalid compilation unit name")?
                 .and_then(|attr| attr.string_value(debug_str));
 
             Unit {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -433,12 +433,20 @@ impl<'input, Endian> Unit<'input, Endian>
...
-                Some(gimli::AttributeValue::DebugLineRef(offset)) => offset,
+                Ok(Some(gimli::AttributeValue::DebugLineRef(offset))) => offset,
+                Err(e) => {
+                    return Err(Error::from(ErrorKind::Gimli(e)))
+                        .chain_err(|| "invalid compilation unit statement list")
+                }
                 _ => return Ok(None),
             };
             let comp_dir = entry.attr(gimli::DW_AT_comp_dir)
+                .map_err(|e| Error::from(ErrorKind::Gimli(e)))
```
philipc: Why do you need the `map_err`? The code compiles without it, and you haven't converted any other gimli errors to the `Gimli` variant prior to this commit, so it seems inconsistent.

jonhoo: Oh, I guess `chain_err` is actually defined for anything that implements `Into<Error>` or something. The `map_err` here is thus redundant, though it does express what is happening internally. I'd be happy to make another PR that removes them if you feel like that'd be better. I don't think it can actually be removed in all the `map_err` cases introduced by this PR though.

src/lib.rs
```diff
@@ -433,12 +433,20 @@ impl<'input, Endian> Unit<'input, Endian>
 
             // Extract source file and line information about the compilation unit
             let line_offset = match entry.attr_value(gimli::DW_AT_stmt_list) {
-                Some(gimli::AttributeValue::DebugLineRef(offset)) => offset,
+                Ok(Some(gimli::AttributeValue::DebugLineRef(offset))) => offset,
+                Err(e) => {
+                    return Err(Error::from(ErrorKind::Gimli(e)))
+                        .chain_err(|| "invalid compilation unit statement list")
```
philipc: Yep, but when I read an error chain that says "invalid compilation unit statement list" I expect that to mean that the statement list is invalid, but it may not be. I'm not sure how to reword it to be concise and accurate. It's a minor point though, and maybe not worth changing, what's there is certainly good enough for debugging.

jonhoo: You're right, it could probably be phrased better. Maybe "invalid attributed encountered while looking up compilation unit statement list"?

src/lib.rs
philipc: None of the `map_err` cases are needed. They do change the error that is returned though. Something like:

    Error(Msg("test"), State { next_error: Some(Error(Gimli(BadUnsignedLeb128), State { next_error: None, backtrace: None })), backtrace: None }

vs

    Error(Msg("test"), State { next_error: Some(BadUnsignedLeb128), backtrace: None })

I'm not sure that the extra `Error(Gimli(..))` provides any value though, because it is behind a trait object pointer so you can't match on it. I haven't read much about matching on error chains though, so was hoping you had a reason for the `map_err` that I could learn from.

src/lib.rs
philipc: The more I think about it, the effort would be better spent improving gimli's errors instead to tell you exactly where the error occurred.

src/lib.rs
jonhoo: Ah, something different is actually going on from what I thought. The former is probably the error format we want. The latter just happens to also be a valid `Result` due to the `Error` wrapping with the `Msg`. The difference between the two is that the former also includes information about where the error came from (namely, that it came from `gimli`), which lets users of the API get a better sense of the true origin of errors. If you look at the defined `error_chain!` towards the top of `src/lib.rs`, you'll see that there's a separate `ErrorKind` used to wrap `gimli` errors, and I think we should make use of it. The reason they both work is that `next_error` is a `Option<Box<error::Error>>`, but `error-chain` encourages you to explicitly type all foreign errors to preserve their origin chain.

PR #16, opened 2016-12-11
Minor: Test file and line equality simultaneously

jonhoo:
Previously, the test cases checked first the file, and then the line number. This meant that if the line number diverged, the file in question was not printed. This PR changes this behavior such that file and line number are checked in a single `assert_eq!`.

#15 should be merged prior to this, and then this should be rebased.


jonhoo:
@fitzgen rebased.

fitzgen:
Thanks!

PR #19, opened 2016-12-13
Dynamic skiplist interval

jonhoo:
Now that gimli-rs/gimli#168 has landed, we can implement the heuristic proposed by @fitzgen in https://github.com/gimli-rs/addr2line/pull/14#issuecomment-266125440 to dynamically determine the interval used by the binary search skiplist introduced in #14 (which uses a hardcoded interval of 100 rows). In particular, this PR approximates `sqrt(#rows)` as suggested by @fitzgen in https://github.com/gimli-rs/addr2line/issues/2#issuecomment-265636994.

This PR also disables the skiplist for units with very few (estimated) rows, which causes some memory savings, at the cost of a minor runtime increase. We can discuss whether we want this behavior, and if we do, what the threshold is.

Performance comparison (`master` is without binary search, `static` is #14, `dynamic` is this PR):
time
memory

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,7 +143,6 @@ struct DebugInfo<'object, Endian>
 {
     debug_line: gimli::DebugLine<'object, Endian>,
     units: Vec<Unit<'object, Endian>>,
-    cache_every: usize,
 }
 
 impl Mapping {
@@ -257,7 +256,6 @@ impl<'object, Endian> DebugInfo<'object, Endian>
 
         Ok(DebugInfo {
             debug_line: debug_line,
-            cache_every: 100,
             units: units,
         })
     }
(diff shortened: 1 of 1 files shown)


fitzgen:
So if I'm reading this right, it looks like `sqrt(estimate(n))` incurs slightly more running time and slightly less memory usage?

jonhoo:
I suspect the difference in runtime and memory is actually caused by the fact that the skiplist is not used for units with fewer than 100 rows (estimated).

jonhoo:
Also, just for future reference, here's the empirical data from three different files I used for the 5.5 heuristic:
```
#bytes  #lines
32991   6350
189490  36421
2436    449
570     100
5159    1017
47      14
-       -
6061    1052
1982    325
4349    726
3226    566
1214    212
1566    334
25415   4595
11083   1886
820     179
1983    317
913     131
787     168
4221    713
2050    343
2951    1126
1429    526
1189    404
15917   2455
17582   2608
436     94
2384    418
-       -
96020   16162
105     65
43      18
39      15
531     300
51      20
163     94
2090    1336
72      30
624     141
9691    1790
228     32
180     20
1572    293
47196   8436
```

PR #21, opened 2016-12-14
Add benchmark comparison against other addr2line implementations

jonhoo:
Following suggestions [on Reddit](https://www.reddit.com/r/rust/comments/5i4yms/the_rust_and_gimli_port_of_addr2line_is_now_an/db5dpt9/), this PR adds `eu-addr2line` from elfutils and `llvm-symbolizer` from llvm to the baseline comparisons in our benchmark. It also cleans up the plotting scripts a little more and enhances the resulting graphs.

Current benchmarking results (i.e., not with #19):
time
memory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 *.png
+tmp.*
--- a/bench.plot.r
+++ b/bench.plot.r
@@ -1,17 +1,21 @@
 v <- read.table(file("stdin"))
 t <- data.frame(prog=v[,1], funcs=(v[,2]=="func"), time=v[,3], mem=v[,4])
 
+t$funcs[t$funcs == TRUE] <- "With functions"
+t$funcs[t$funcs == FALSE] <- "File/line only"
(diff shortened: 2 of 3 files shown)


jonhoo:
I've also wondered whether we should show a scatterplot instead to also indicate variance, but it's sort of tricky given that we can't trivially measure per-address lookup time.

fitzgen:
Any idea why the OSX builds failed with `!` ?

cuviper:
AFAICS, this addr2line doesn't unwind inlines (`-i`) yet, right?  When you do, that will be good to benchmark too, as it requires more walking of the DIE trees.

jonhoo:
@cuviper we actually walk the full tree already to resolve the innermost linkage name, so I suspect the difference won't be all that great. But you are completely right that we don't currently support `-i`, so it's hard to say for sure.

jonhoo:
Note to self + @fitzgen: we should probably incorporate these two plots into the `README`.

PR #24, opened 2017-01-18
Expose symbolication errors at creation time

jonhoo:
Now that https://github.com/Kimundi/owning-ref-rs/issues/19 has been resolved with `ObjectHandle` gaining a `try_new` method, we can get rid of the nasty `.expect()` in `Mapping::new_inner`, which would panic if the debug symbols for a given binary were invalid. Now, the error is instead exposed directly to the caller through the returned `Result`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,22 +173,18 @@ impl Mapping {
         let file = memmap::Mmap::open_path(file_path, memmap::Protection::Read)
             .map_err(|e| ErrorKind::BadPath(e))?;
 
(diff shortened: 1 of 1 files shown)


philipc:
Looks good.

Unrelated to this, kcov is no longer working. Not sure if the libbfd error is cause, and I don't know why it isn't finding libbfd because that should be in binutils-dev.

jonhoo:
@philipc yeah, I noticed that too, but don't really have an explanation either :/

PR #28, opened 2017-01-23
Add Phillip and Jon as authors

jonhoo:
Not terribly important, but since we've been doing a pass on `Cargo.toml` anyway..

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-authors = ["Nick Fitzgerald <fitzgen@gmail.com>"]
+authors = ["Nick Fitzgerald <fitzgen@gmail.com>", "Philip Craig <philipjcraig@gmail.com>", "Jon Gjengset <jon@thesquareplanet.com>"]
 description = "A cross-platform `addr2line` clone written in Rust, using `gimli`"
 documentation = "https://docs.rs/addr2line"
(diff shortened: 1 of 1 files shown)


jonhoo:
@philipc The coveralls issue is apparently known: https://github.com/huonw/travis-cargo/issues/60

philipc:
That's actually a different coveralls issue, and our `coverage` script is a workaround for that one.

jonhoo:
Ah, ours is actually that `kcov` itself [segfaults](https://travis-ci.org/gimli-rs/addr2line/jobs/194353361#L1084). I wonder why...

Also, presumably once a fix for the issue above lands, we can get rid of our `coverage` script?

philipc:
Yeah, the current issue is that I think `kcov` is segfaulting because it needs the --verify option, but it's missing the --verify option because it can't find libbfd, and it can't find libbfd because something has changed in travis. I have no idea why it can't find libbfd, because our `.travis.yml` specifies `binutils-dev`, and on my ubuntu version that contains `/usr/lib/x86_64-linux-gnu/libbfd.so`.

We can get rid of our `coverage` script once a fix lands, but travis-cargo seems unmaintained. I don't think huonw works on rust anymore (and possibly isn't allowed to).

fitzgen:
Maybe if we specify LD_LIBRARY_PATH it can find the so? Although it
certainly seems like that path should be searched by default...

PR #31, opened 2017-02-18
Add support for C++ symbol demangling

### LINKED ISSUE #17: Use `cpp_demangle` crate for demangling C++ symbols, once its ready
Ongoing WIP over here: https://github.com/fitzgen/cpp_demangle

fitzgen:
The `cpp_demangle` crate doesn't match libiberty 100% yet, but it is in a good enough place that we can get some utility out of it here.

Fixes #17 

r? @jonhoo

### DIFF
--- a/src/bin/addr2line.rs
+++ b/src/bin/addr2line.rs
@@ -1,12 +1,8 @@
 #[macro_use]
 extern crate clap;
 extern crate addr2line;
-extern crate rustc_demangle;
-use rustc_demangle::demangle;
 
 use clap::{App, Arg};
-
-use std::path;
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 const POSTSCRIPT: &'static str = "\
@@ -25,8 +21,8 @@ question marks in their place. If the line number can not be
 determined, addr2line will print 0.
 ";
 
-fn main() {
-    let matches = App::new("addr2line")
+fn get_matches() -> clap::ArgMatches<'static> {
+    App::new("addr2line")
         .version(crate_version!())
         .about("Convert addresses into line number/file name pairs.\n\
                 If no addresses are specified on the command line, they will be read from stdin.")
@@ -56,16 +52,41 @@ fn main() {
             .multiple(true)
             .index(1))
         .after_help(POSTSCRIPT)
-        .get_matches();
+        .get_matches()
+}
+
+#[cfg(any(feature = "rustc-demangle", feature = "cpp_demangle"))]
+fn get_options(matches: &clap::ArgMatches) -> addr2line::Options {
+    let mut opts = addr2line::Options::default();
+
+    if matches.is_present("functions") {
+        opts = opts.with_functions();
+    }
+    if matches.is_present("demangle") {
+        opts = opts.with_demangling();
+    }
+
+    opts
+}
+
+#[cfg(not(any(feature = "rustc-demangle", feature = "cpp_demangle")))]
+fn get_options(matches: &clap::ArgMatches) -> addr2line::Options {
+    let mut opts = addr2line::Options::default();
+
+    if matches.is_present("functions") {
+        opts = opts.with_functions();
+    }
+
+    opts
+}
+
+fn main() {
+    let matches = get_matches();
+    let opts = get_options(&matches);
 
     let exe = matches.value_of("executable").unwrap_or("./a.out");
     let show_funcs = matches.is_present("functions");
-
-    let debug = if show_funcs {
-        addr2line::Mapping::with_functions(path::Path::new(&exe))
-    } else {
-        addr2line::Mapping::new(path::Path::new(&exe))
-    };
+    let debug = opts.build(exe);
 
     if let Err(e) = debug {
         println!("addr2line: {:?}", e);
@@ -74,7 +95,6 @@ fn main() {
     let debug = debug.unwrap();
 
     let show_addrs = matches.is_present("addresses");
-    let do_demangle = matches.is_present("demangle");
     let one = |addr: &str| {
         let addr = parse_uint_from_hex_string(addr);
         if show_addrs {
@@ -90,13 +110,7 @@ fn main() {
         if let Some((file, lineno, func)) = loc {
             if show_funcs {
                 use std::borrow::Cow;
-                println!("{}",
-                         func.map(|func| if do_demangle {
-                                 Cow::Owned(demangle(&*func).to_string())
-                             } else {
-                                 func
-                             })
-                             .unwrap_or(Cow::Borrowed("??")));
+                println!("{}", func.unwrap_or(Cow::Borrowed("??")));
             }
             println!("{}:{}",
                      file.to_string_lossy(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,13 @@ extern crate memmap;
 extern crate object;
 extern crate owning_ref;
 extern crate fallible_iterator;
+
+#[cfg(feature = "rustc-demangle")]
+extern crate rustc_demangle;
+
+#[cfg(feature = "cpp_demangle")]
+extern crate cpp_demangle;
+
 #[macro_use]
 extern crate error_chain;
 
@@ -115,6 +122,60 @@ mod errors {
 }
 pub use errors::*;
 
+/// A builder for configuring a `Mapping`.
+///
+/// ```
+/// # fn foo() -> addr2line::Result<()> {
+/// use addr2line::Options;
+///
+/// let mapping = Options::default()
+///     .with_functions()
+///     .build("path/to/some/executable")?;
+/// # let _ = mapping;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// If `with_functions` is not added to these options, then `Mapping::locate()`
+/// will always return `None` for the function name.
+///
+/// Likewise, if `with_demangling` is not added to these options, then all
+/// function names (if any) returned by `Mapping::locate()` will not be
+/// demangled.
+#[derive(Clone, Copy, Default)]
+pub struct Options {
+    with_functions: bool,
+    with_demangling: bool,
+}
+
+impl Options {
+    /// Make the `Mapping` also include debug information for functions,
+    /// enabling `Mapping::locate()` to also indicate what function a given
+    /// address appears in. This comes at some parsing and lookup cost.
+    pub fn with_functions(mut self) -> Self {
+        self.with_functions = true;
+        self
+    }
+
+    /// Make the `Mapping` attempt to demangle Rust and/or C++ symbols. This
+    /// option implies `with_functions`.
+    #[cfg(any(feature = "cpp_demangle", feature = "rustc-demangle"))]
+    pub fn with_demangling(mut self) -> Self {
+        self.with_demangling = true;
+        self.with_functions()
+    }
+
+    /// Finish configuration and build the `Mapping`.
+    ///
+    /// The target file will be memmap'd, and then `gimli` is used to parse out
+    /// the necessary debug symbols, without copying data when possible.
+    pub fn build<P>(self, file_path: P) -> Result<Mapping>
+        where P: AsRef<path::Path>
+    {
+        Mapping::new_inner(file_path.as_ref(), self)
+    }
+}
+
 /// A `Mapping` locates and maintains the state necessary to perform address to line translation.
 ///
 /// Constructing a `Mapping` is somewhat costly, so users should aim to re-use created `Mapping`s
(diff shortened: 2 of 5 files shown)


fitzgen:
Fixes #17

src/lib.rs
```diff
@@ -454,7 +489,21 @@ impl<'object, Endian> DebugInfo<'object, Endian>
                 // matches us better. We need to keep going until we find the best one.
             }
 
-            return Ok(Some((path, line, func.map(|u| u.0.name.to_string_lossy()))));
+            let func = func.map(|u| {
+                if self.with_demangling {
+                    // First try demangling as a C++ symbol, then fall back to
+                    // Rust symbols.
```
philipc: Should we be using the compilation unit's `DW_AT_language` attribute to determine which demangling to use instead?

src/bin/addr2line.rs
```diff
@@ -1,13 +1,9 @@
 #[macro_use]
 extern crate clap;
 extern crate addr2line;
-extern crate rustc_demangle;
```
jonhoo: Can presumably also be removed from `Cargo.toml`?

src/bin/addr2line.rs
```diff
@@ -61,11 +57,16 @@ fn main() {
...
     let show_funcs = matches.is_present("functions");
 
-    let debug = if show_funcs {
-        addr2line::Mapping::with_functions(path::Path::new(&exe))
-    } else {
-        addr2line::Mapping::new(path::Path::new(&exe))
-    };
+    let mut opts = addr2line::Options::default();
+
+    if show_funcs {
```
jonhoo: Just for consistency, I think I'd prefer this to be `matches.is_present("functions")`, and then to do the `let show_funcs` along with `let show_addrs` below (i.e., right above the `let one =`). It does mean that we repeat `is_present("functions")`, but the code reads nicer.

src/bin/addr2line.rs
```diff
@@ -90,13 +90,7 @@ fn main() {
...
             if show_funcs {
                 use std::borrow::Cow;
-                println!("{}",
-                         func.map(|func| if do_demangle {
-                                 Cow::Owned(demangle(&*func).to_string())
-                             } else {
-                                 func
-                             })
-                             .unwrap_or(Cow::Borrowed("??")));
+                println!("{}", func.unwrap_or(Cow::Borrowed("??")));
```
jonhoo: Out of curiosity, what's the rationale for moving demangling into the `addr2line` internals? I kind of liked the fact that it was entirely external to the `addr2line` lookup logic.

src/lib.rs
```diff
@@ -22,17 +22,20 @@
...
 extern crate owning_ref;
 extern crate fallible_iterator;
+extern crate rustc_demangle;
 #[macro_use]
 extern crate error_chain;
 
 use object::Object;
 use owning_ref::OwningHandle;
 use fallible_iterator::FallibleIterator;
+use rustc_demangle::demangle;
```
jonhoo: We should either do a `use cpp_demangle::Symbol` here, or we should remove the `use` for `rustc_demangle`.

src/lib.rs
```diff
@@ -115,6 +118,51 @@ mod errors {
...
+/// ```
+/// # fn foo() -> addr2line::Result<()> {
+/// use addr2line::Options;
+///
+/// let mapping = Options::default()
+///     .with_functions()
+///     .build("path/to/some/executable")?;
+/// # let _ = mapping;
+/// # }
+/// ```
```
jonhoo: I'd like this to include a statement along the lines of "if `with_functions` is not added, `locate()` will always return `None` for the function".

src/lib.rs
```diff
@@ -143,33 +191,21 @@ struct DebugInfo<'object, Endian>
 {
     debug_line: gimli::DebugLine<'object, Endian>,
     units: Vec<Unit<'object, Endian>>,
+    with_demangling: bool,
```
jonhoo: The more I see this, the more I want demangling to stay outside the internals of the lookup code.  For example, it would be entirely sufficient for `Mapping::locate` to do the demangling. And this would avoid the need to propagate down the `with_demangling` bool.

src/lib.rs
```diff
@@ -244,7 +278,7 @@ impl<'object, Endian> DebugInfo<'object, Endian>
                                    &debug_line,
                                    &debug_str,
                                    &header,
-                                   with_functions);
+                                   opts.with_functions);
```
jonhoo: To be consistent, this should probably be `opts`, even if we don't need the `with_demangling` information.

src/lib.rs
```diff
@@ -254,6 +288,7 @@ impl<'object, Endian> DebugInfo<'object, Endian>
         Ok(DebugInfo {
             debug_line: debug_line,
             units: units,
+            with_demangling: opts.with_demangling,
```
jonhoo: Same as above, this should be a full `opts` (unless we decide to hoist the demangling code).

src/lib.rs
jonhoo: Or I guess technically `DebugInfo::locate`, so that we can take `DW_AT_language` into account as @philipc suggested.

jonhoo: This should probably carry on the full `Options`; we may want to add things to it later.

src/lib.rs
```diff
@@ -115,6 +118,51 @@ mod errors {
...
+    pub fn with_demangling(mut self) -> Self {
+        self.with_demangling = true;
+        self.with_functions()
+    }
+
+    /// Finish configuration and build the `Mapping`.
+    ///
+    /// The target file will be memmap'd, and then `gimli` is used to parse out
+    /// the necessary debug symbols, without copying data when possible.
+    pub fn build<P>(self, file_path: P) -> Result<Mapping>
```
jonhoo: One option here would be to have `Options::new`, which requires `file_path` (and not implement `Default`), and then implement `Into<Mapping>` for `Options`.

src/lib.rs
```diff
@@ -454,7 +489,21 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                    // Rust symbols.
+                    if let Ok(sym) = cpp_demangle::Symbol::new(u.0.name.to_bytes()) {
+                        Cow::from(format!("{}", sym))
+                    } else {
+                        Cow::from(format!("{}", demangle(u.0.name.to_string_lossy().as_ref())))
+                    }
+                } else {
+                    u.0.name.to_string_lossy()
+                }
+            });
```
jonhoo: So, I've been going back and forth on this now a bunch (you may have seen some comments appear and then disappear). I really dislike that demangling is pushed this far into `addr2line`, when it can be almost entirely externally. Admittedly, if we want to use `DW_AT_language`, we may have to do it here, but if not, this could be done in `Mapping::locate`. At the very least I'd like to see the closure above moved into a function.

jonhoo:
Left a bunch of comments inline. The biggest hesitation I have about this change is that it adds a dependency pretty deep inside `addr2line` that doesn't *really* need to be there. Previously, demangling was done entirely after the lookup was performed, which made it orthogonal to the rest of the code. In fact, you could (and the old `src/main.rs` did) do demangling completely outside `addr2line`. Maybe it is necessary to push it this deep (e.g., if we want to use `DW_AT_language` as @philipc suggested), but I'm somewhat skeptical...

fitzgen:
> The biggest hesitation I have about this change is that it adds a dependency pretty deep inside addr2line that doesn't really need to be there. Previously, demangling was done entirely after the lookup was performed, which made it orthogonal to the rest of the code. In fact, you could (and the old src/main.rs did) do demangling completely outside addr2line.

I'd be happy to make it an optional (but on-by-default) dependency. That shouldn't be too hard.

Whether we even support demangling or not in the library is a bit of a philosophical question: batteries included or not? I'm leaning towards batteries included: why force library users to demangle themselves, reproducing basically exactly the code that would be in `src/main.rs`?

> Maybe it is necessary to push it this deep (e.g., if we want to use DW_AT_language as @philipc suggested), but I'm somewhat skeptical...

If we use DW_AT_language, then it will be necessary, because you have to have access to the CU to get that info, and additionally we can't assume all CUs have the same DW_AT_language (eg linking C++ and Rust together in one executable).

src/lib.rs
fitzgen: It doesn't seem to me that either option is better than the other, so unless you have strong opinions, then I'll leave it as is.

fitzgen:
@philipc @jonhoo 

Ok, I think I've addressed both of your comments now.

* Both `cpp_demangle` and `rustc-demangle` are optional dependencies now, and both are enabled by default.

* We use `DW_AT_lang` to choose which demangling to perform.

* Various nitpick fixes from @jonhoo's review.

src/bin/addr2line.rs
```diff
@@ -56,16 +52,41 @@ fn main() {
...
+#[cfg(any(feature = "rustc-demangle", feature = "cpp_demangle"))]
+fn get_options(matches: &clap::ArgMatches) -> addr2line::Options {
+    let mut opts = addr2line::Options::default();
+
+    if matches.is_present("functions") {
+        opts = opts.with_functions();
+    }
+    if matches.is_present("demangle") {
+        opts = opts.with_demangling();
+    }
```
jonhoo: This would be a lot more concise with `if cfg!`, though I'm unsure whether that would compile? In theory the compiler could be clever enough to let it work even when the features are off.

src/bin/addr2line.rs
```diff
@@ -56,16 +52,41 @@ fn main() {
...
+
+    if matches.is_present("functions") {
+        opts = opts.with_functions();
+    }
+
+    opts
+}
+
+fn main() {
+    let matches = get_matches();
```
jonhoo: Any particular reason you wanted to move out `get_matches` even though that doesn't need the `#[cfg]` guard?

src/bin/addr2line.rs
fitzgen: Unfortunately, that doesn't work here. Annoying. One of the few situations where the c preprocessor is nicer to use.

src/lib.rs
```diff
@@ -454,12 +502,53 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+            let func = func.map(|u| {
+                if unit.language.is_some() {
+                    debug_assert!(self.opts.with_demangling,
+                                  "We shouldn't even bother finding the DW_AT_language if we \
+                                   aren't demangling");
+                }
+                match unit.language {
+                    Some(gimli::DW_LANG_C_plus_plus) |
+                    Some(gimli::DW_LANG_C_plus_plus_03) |
+                    Some(gimli::DW_LANG_C_plus_plus_11) => demangle_cpp_symbol(u.0.name),
```
jonhoo: Does `C` not use mostly the same mangling method?

src/lib.rs
```diff
@@ -454,12 +502,53 @@ impl<'object, Endian> DebugInfo<'object, Endian>
...
+                    Some(gimli::DW_LANG_Rust) => demangle_rust_symbol(u.0.name),
+                    _ => u.0.name.to_string_lossy(),
```
jonhoo: It'd be nice if we could report this case to the user somehow. If they ask for demangling, and we don't support the language that's in use, we should report it. I'm not entirely sure what the best way to do so is though.

Also, it *could* be that some demangling is better than no demangling, even for unknown languages. Would it be better to fall back to `cpp` here?

jonhoo:
Yeah, I'm coming around to the idea of demangling being a built-in. It depends on whether we expect people who use `addr2line` *as a library* will care about function names, but I think you might be right that that's likely to be the case.

src/bin/addr2line.rs
jonhoo: And we can't `#[cfg]` annotate an `if`?

src/lib.rs
fitzgen: C doesn't generally use any mangling :)

src/bin/addr2line.rs
fitzgen: Nope :(

https://is.gd/npNRHi

src/bin/addr2line.rs
fitzgen: Er, nope, https://is.gd/npNRHi

src/bin/addr2line.rs
fitzgen: I just felt that the function was getting huge. I think it is better factored this way.

src/bin/addr2line.rs
jonhoo: Too bad. Oh well, I guess we're stuck with having it this way for now.

src/lib.rs
fitzgen: Yeah I'm not sure what the best way to do that would be. Maybe if we added `env_logger` or something, but I don't think that is necessarily a good idea...

fitzgen:
OK, I'm going to merge this, and we can make any more changes in follow ups.

Issue #30, opened 2017-02-15
There exist DWARF files in the wild where a program entry's ranges are not contained within the CU


khuey:
Using the separate debug info for libpthread.so.0 on Xenial.

>~/dev/addr2line/target/debug/addr2line -f -e /usr/lib/debug/.build-id/4d/7f52f335dc9665c2dcf308ce6514a6ae86dede.debug 5555
>
>thread 'main' panicked at 'assertion failed: unit.ranges.iter().any(|range| range.begin <= r.begin && range.end >= r.end)', src/lib.rs:597

The DWARF contains an inline subprogram that is not within the CU's declared ranges.

fitzgen:
Nice find

jonhoo:
That check *should* only be present in debug builds. In a release build, it wouldn't panic, but the inlined subprogram would not be found when resolving.

philipc:
debug asserts should still only check for things that should be logically impossible. The only reason they aren't also checked in release builds is for performance (and the documentation for `debug_assert` states that it should only be used instead of `assert` if the performance difference is measurable).

jonhoo:
From our earlier discussions about this when it was originally introduced, I would argue that this should never happen, and therefore a debug assert is (or rather, was) reasonable. Now that we've seen a real-world binary with this invariant broken, I agree, that should change. One option is to just remove the assert, and leave the code as is otherwise. Having a subprogram that is not within the CU's declared ranges seems like a bug to me, and it is somewhat reasonable to simply ignore such subprograms. The other option would be more drastic — to fix this we'd need to rewrite the code quite a bit, since it currently *assumes* that this invariant holds in order to do the search efficiently.

I wonder what the other symbol-finding tools do for this particular subprogram. @khuey can you try doing a lookup for an address in the range in question using `addr2line` from `binutils` (or some other place), and compare it to the name resolved by our `addr2line` when compiled in release mode?

khuey:
>khuey@minbar:/home/khuey/dev/addr2line$ ./target/release/addr2line -f -e /usr/lib/debug/.build-id/4d/7f52f335dc9665c2dcf308ce6514a6ae86dede.debug 226
>??
>??:?
>khuey@minbar:/home/khuey/dev/addr2line$ addr2line -f -e /usr/lib/debug/.build-id/4d/7f52f335dc9665c2dcf308ce6514a6ae86dede.debug 226??
>??:0

jonhoo:
[...]

PR #32, opened 2017-02-26
Remove invalid debug assertion

jonhoo:
Following discussion in #30, it seems clear that asserting that all of a compilation unit's child units are contained within it is wrong. It happens in the real world, and is thus not *impossible*.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -587,14 +587,19 @@ impl<'input, Endian> Unit<'input, Endian>
                 continue;
             }
 
-            if cfg!(debug_assertions) {
-                // All programs should be fully contained within their compilation unit
-                for r in &ranges {
-                    assert!(unit.ranges
-                        .iter()
-                        .any(|range| range.begin <= r.begin && range.end >= r.end));
-                }
-            }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -587,14 +587,19 @@ impl<'input, Endian> Unit<'input, Endian>
...
+            // contained within compilation unit's range. That is
+            //
+            //   ∀r ∈ cu.ranges (cu.begin <= r <= cu.end)
+            //
+            // It turns out that this is not true: https://github.com/gimli-rs/addr2line/issues/30.
+            // In such programs, we will fail to resolve addresses located in units whose range
+            // fall outside the range of the containing compilation unit. This is not a hard error,
+            // since we can still generally resolve addresses, and tools like binutils' addr2line
+            // exhibit the same behavior, but it is something we should aim to eventually work
+            // around. Hence: TODO
```
fitzgen: > Hence: TODO

😆

jonhoo:
Note that we *might* want to revert this if @philipc's comment bears our and fixes the issue that caused the assert to be triggered in the first place.

PR #33, opened 2017-02-28
Use base address of compilation unit for range lists

### LINKED ISSUE #30: There exist DWARF files in the wild where a program entry's ranges are not contained within the CU

philipc:
When parsing range lists, we must use the base address of the
compilation unit, rather than that of the entry referring to the
range list.

Also try parse to DW_AT_entry_pc if DW_AT_low_pc is not present.
This is for compatibility with GNU tools.

Fixes #30

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -558,6 +558,7 @@ struct Unit<'input, Endian>
 
     cache_every: Option<usize>,
     address_size: u8,
+    base_address: u64,
     ranges: Vec<gimli::Range>,
     line_offset: gimli::DebugLineOffset,
     comp_dir: Option<&'input std::ffi::CStr>,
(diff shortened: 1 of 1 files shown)


jonhoo:
LGTM. r+

PR #36, opened 2017-03-14
Handle DW_AT_MIPS_linkage_name and DW_AT_specification

### LINKED ISSUE #35: addr2line -f fails with "InvalidDebugSymbols(SubroutineMissingName(0, 100459))"
Way to reproduce:

    wget http://slonopotamus.org/libUE4Editor-Core.so
    addr2line -f -e libUE4Editor-Core.so 0x2c09b2

Expected output (as produced by `addr2line` from GNU binutils and gdb, and I know this is correct result):

    _Z20PlatformCrashHandleriP9siginfo_tPv
    /home/marat/production/ue4/Engine/Source/Runtime/Core/Private/Linux/LinuxPlatformCrashContext.cpp:595

Actual output:

    addr2line: Error(Msg("failed to analyze debug information"), State { next_error: Some(Error(Msg("encountered invalid compilation unit"), State { next_error: Some(Error(InvalidDebugSymbols(SubroutineMissingName(0, 100459)), State { next_error: None, backtrace: None })), backtrace: None })), backtrace: None })

Note that running without `-f` switch properly resolves filename/lineno, it's only function name that fails to resolve.

Funny facts:

1. `addr2line` from `elfutils` [works in the opposite way][1]: it resolves function name but fails with filename/lineno.
2. `addr2line` from `elftoolchain` [segfaults][2] when trying to resolve function name, and [gives wrong filename/lineno][3].
3. `addr2line` from FreeBSD [gives wrong filename/lineno and fails to find function name][4]

So, among five `addr2line` implementations, there are no two that behave the same way.

[1]: https://sourceware.org/bugzilla/show_bug.cgi?id=21247
[2]: https://sourceforge.net/p/elftoolchain/tickets/546/
[3]: https://sourceforge.net/p/elftoolchain/tickets/545/
[4]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=217736

philipc:
Also some noise from running newer rustfmt.

Fixes #35

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -757,6 +757,12 @@ impl<'input, Endian> Unit<'input, Endian>
                .and_then(|attr| attr.string_value(debug_str)) {
             return Ok(Some(name));
         }
+        if let Some(name) = entry.attr(gimli::DW_AT_MIPS_linkage_name)
+               .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+               .chain_err(|| "invalid subprogram linkage name")?
+               .and_then(|attr| attr.string_value(debug_str)) {
+            return Ok(Some(name));
+        }
 
         // Linked name is not available, so fall back to just plain old name, if that's available.
         if let Some(name) = entry.attr(gimli::DW_AT_name)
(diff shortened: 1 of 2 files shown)


fitzgen:
Perhaps we should do a new point release as well?

jonhoo:
@philipc Nice!
@fitzgen Yeah, that seems reasonable.

Issue #38, opened 2017-07-07
Mapping libxul.so on Linux panics in xmas_elf::ElfFile::get_string due to bad slice


mstange:
I'm not having any luck using addr2line to symbolicate a Firefox libxul.so on Linux.

My code:
```rust
extern crate addr2line;

fn main() {
    let mapping = addr2line::Options::default()
        .build(
            "/home/mstange/code/mozilla/obj-x86_64-pc-linux-gnu/dist/bin/libxul.so",
        ).expect("Mapping libxul.so failed");
    println!("0x33e4812: {:?}", mapping.locate(0x33e4812));
}
```

Output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/addr2line-libxul`
thread 'main' panicked at 'slice index starts at 77309416265 but ends at 1434107856', /checkout/src/libcore/slice/mod.rs:741:4
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:380
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:611
   5: std::panicking::begin_panic_new
             at /checkout/src/libstd/panicking.rs:553
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:521
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:92
   9: core::slice::slice_index_order_fail
             at /checkout/src/libcore/slice/mod.rs:741
  10: <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index
             at /checkout/src/libcore/slice/mod.rs:864
  11: <core::ops::range::RangeFrom<usize> as core::slice::SliceIndex<[T]>>::index
             at /checkout/src/libcore/slice/mod.rs:947
[...]

PR #39, opened 2017-07-07
Update deps

fitzgen:
r? @philipc or @jonhoo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "Apache-2.0/MIT"
 name = "addr2line"
 readme = "./README.md"
 repository = "https://github.com/gimli-rs/addr2line"
-version = "0.3.0"
+version = "0.4.0"
 build = "build.rs"
 
 [badges]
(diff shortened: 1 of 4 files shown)


jonhoo:
@fitzgen any chance you could submit the rustfmt changes separately? unless the dependency bump *only* changes `Cargo.toml`...

philipc:
@jonhoo The dependency bumps change more than `Cargo.toml`. If you're just wanting to see what those changes were, then they're already in separate commits that you can review individually.

fitzgen:
> this doesn't fix everything in #38 though does it?

Not sure; don't have concrete steps to reproduce or test case, so can't verify myself.

> @fitzgen any chance you could submit the rustfmt changes separately? unless the dependency bump only changes Cargo.toml...

As @philipc said, they're separate commits, so if you click on the "commits" tab here (in between "Discussion" and "Files Changed", you can see each commit's change in isolation :)

fitzgen:
Published 0.4.0 on crates.io

PR #45, opened 2017-08-14
Add BufferMapping

### LINKED ISSUE #37: Support byte based API ?
I would like to use this in an application but it seems the only entry point in the API is a path based one. 

Do you plan on supporting an `AsRef<[u8]>` or  a simple &[u8] entry point ?

As it stands now the API as is would essentially double memory usage for a consumer which has already loaded / obtained the bytes in some lawful (or unlawful) manner, as it appears to memmap the binary/bytes. 

Thanks and awesome crate!

philipc:
Fixes #37 

I'm open to bikeshedding for the name `BufferMapping`.

cc @m4b

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ fallible-iterator = "0.1.3"
 clap = "2.19.1"
 gimli = "0.14.0"
 memmap = "0.5.0"
-object = "0.4.0"
+object = "0.4.1"
 owning_ref = "0.2.4"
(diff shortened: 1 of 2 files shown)


m4b:
Well this is awesome !

Iiuc, to use with pre allocated bytes, just BufferMapping::new and then locate for vm addresses I want to investigate yes ?

philipc:
Yes, or if you want to use non-default options then use `Options::build_from_buffer`.

Issue #29, opened 2017-01-23
Assumption of monotonically increasing addresses in line number programs is wrong


khuey:
in practice, not just in theory.  See https://gist.github.com/khuey/5b0a7d0cc68360e85cd31c67b0c0b0e7.

philipc:
@khuey Did you end up using the sequences API that was added to gimli? I looked at changing addr2line to use it, but it turned out much slower than I had hoped, because some sequences can be very long.

So I'm now thinking that we need to keep the existing skiplist but:
- force a skiplist entry at the start of sequences in addition to every nth row
- fully populate and sort the skiplist, instead of lazily reading rows

It'll require more memory than only caching the start of sequences, but I don't have any better ideas.

khuey:
I am using it.  Performance hasn't been a big concern for us but I haven't tried it on something insane like the single-file sqlite.c.

philipc:
Yet another approach is to just cache every row. This has the advantage of being faster at the cost of more memory. We don't need to cache the full row though. This is also what every other library I've looked at does (including libbacktrace). I have nearly completed implementation of this approach (see https://github.com/philipc/addr2line/commit/aa56a3a42f8318b15a63cbec425f9aa0f4c68cc9).

PR #46, opened 2017-08-15
End sequence

fitzgen:
r? @philipc

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -367,10 +367,17 @@ where
             };
 
             // Now, find the last row before a row with a higher address than the one we seek.
-            let mut praddr = 0;
+            let mut prev_row_addr = 0;
             let mut skipseq = false;
             while let Ok(Some((_, &row))) = rows.next_row() {
+                let row_addr = row.address();
+
                 if row.end_sequence() {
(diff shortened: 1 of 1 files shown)


fitzgen:
Beat you to it ;)

This fixes the issues I was seeing in my ongoing work to integrate this crate into backtrace-rs

fitzgen:
We should cut a new release after this and the other PR merge.

philipc:
Approved under the assumption that you've tested and it works. But the logic in that loop is convoluted, and it's not clear to me that there's no more holes in it. So I'll do another PR later today to clean that up.

> We should cut a new release after this and the other PR merge.

I consider #29 to be important to fix.

I have a fix based on converting it to use sequences. It's a lot slower, but correctness is better than speed. Otherwise I can do the change I described in https://github.com/gimli-rs/addr2line/issues/29#issuecomment-321959260, but I haven't started on that yet.

philipc:
Oops, I didn't notice this failed the build.

philipc:
So the failure is due to a bug in binutils addr2line. It tries to be too smart about handling non-increasing addresses within a sequence, and gets it wrong.

fitzgen:
Huh. Tests were passing locally, I guess I have a different version of binutils on my machine.

PR #48, opened 2017-08-17
Handle out of order sequences

### LINKED ISSUE #29: Assumption of monotonically increasing addresses in line number programs is wrong

philipc:
Fixes #29 

Now caches every line number row, instead of using a skiplist.

`locate` API has changed to require `&mut self` (and no longer uses `RwLock` internally). Not sure if this is a problem, but I wanted to get it working correctly before adding complications like that.

Performance is better for both CPU and memory when testing symbolication of 26000 addresses in a 53MB library. Timing from `/usr/bin/time -f '%e\t%M'`:

```
Without functions:
Before: 0.29	32396
After:  0.18	32260

With functions:
Before: 1.89	72580
After:  1.73	72320
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -280,7 +280,7 @@ impl<'input> BufferMapping<'input> {
     pub fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<str>>)>> {
+    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<'input, str>>)>> {
         self.0.locate(addr)
     }
 }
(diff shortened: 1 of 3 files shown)


fitzgen:
> locate API has changed to require &mut self (and no longer uses RwLock internally). Not sure if this is a problem, but I wanted to get it working correctly before adding complications like that.

This seems like the correct decision to me. If people want to share these things across threads, they can provide their own locking wrapping the whole `Mapping`.

src/lib.rs
```diff
@@ -1046,3 +889,140 @@ where
...
+                rows: sequence_rows,
+            });
+        }
+        // Sort so we can binary search.
+        sequences.sort_by(|a, b| a.low_address.cmp(&b.low_address));
+        self.sequences = sequences;
+        self.read_sequences = true;
+    }
+
+    fn locate(&self, address: u64) -> Option<&Row> {
```
fitzgen: We should `debug_assert!` that we've read the sequences as the first thing in this function for future code readers' benefit. I just had to track down where we actually invoke `read_sequences()`, so there's sample size of 1 that this isn't immediately obviously correct, and the assertion would quiet such concerns ;)

philipc: Yep good point. I originally had the `read_sequences()` call here, but it didn't work with the borrow checker.

philipc:
I added a commit that improves the end of sequence handling a bit too.

PR #50, opened 2017-08-18
Bump to version 0.5.0

fitzgen:
Breaking bump because `Mapping::locate` changed from `&self` to `&mut self`.

r? @philipc


philipc:
I just noticed the README needs updating for the `locate` change.

fitzgen:
Pushed a commit fixing the README example.

PR #52, opened 2017-10-02
Various cleanup

philipc:
The integration test was failing because we are returning a filename for the main() wrapper, but binutils doesn't. I changed the integration test to ignore this case.

Also run rustfmt and clippy.


jonhoo:
Ah, looks great! Could you also fix the `kcov` failure on `nightly` (or set `allow_failures` on `nightly`)?

philipc:
> Could you also fix the kcov failure on nightly (or set allow_failures on nightly)?

Yeah, I'll do that in another PR.

PR #53, opened 2017-10-02
Add benchmark

philipc:
The fixture was built from the current source of this crate with:
```
RUSTFLAGS=-g cargo build --release
objcopy --only-keep-debug target/release/addr2line fixtures/addr2line-release
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ readme = "./README.md"
 repository = "https://github.com/gimli-rs/addr2line"
 version = "0.5.0"
 build = "build.rs"
+exclude = ["/benches/*", "/fixtures/*"]
 
 [badges]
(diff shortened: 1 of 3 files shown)


jonhoo:
Hmm, I wonder if we should benchmark `build()` and `lookup()` separately. They do fairly different work, and if one got slower and the other got faster, we'd want to know.

jonhoo:
I also don't know how reliable Travis performance is, but could we maybe add this as a test to `.travis.yml`?

philipc:
> I wonder if we should benchmark build() and lookup() separately.

Good idea. I've added another commit for that. It's a bit hard to get clean individual benchmarks though, due to lazy parsing of line number information.

> I also don't know how reliable Travis performance is, but could we maybe add this as a test to .travis.yml?

It actually already did that, even though there were no benchmarks.

PR #54, opened 2017-10-03
Use cargo-tarpaulin instead of kcov

philipc:
Also stop use travis-cargo because it is unmaintained and no longer provides any benefit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,28 +17,23 @@ rust:
 addons:
   apt:
     packages:
-    - libcurl4-openssl-dev
-    - libelf-dev
-    - libdw-dev
-    - binutils-dev
-
-before_script:
-- pip install -v 'travis-cargo<0.2' --user
-- if [[ -e ~/Library/Python/2.7/bin ]]; then export PATH=~/Library/Python/2.7/bin:$PATH; fi
-- if [[ -e ~/.local/bin ]]; then export PATH=~/.local/bin:$PATH; fi
+    - libssl-dev
 
 script:
--  travis-cargo build -- --no-default-features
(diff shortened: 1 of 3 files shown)


philipc:
Still timed out...
Worked fine when I tested it here though: https://travis-ci.org/philipc/addr2line/builds/282569006

jonhoo:
That's a little disturbing. It really shouldn't take that much longer from one run to the next on the same fixture.. Let me try restarting it.

jonhoo:
@philipc hmm, the re-run also timed out (https://travis-ci.org/gimli-rs/addr2line/jobs/282572460). Tried restarting it a second time, but this suggests a deeper issue..

philipc:
Did some investigating:
- the hang occurs locally too
- I've only seen it hang on the self tests
- when the hang occurs cpu usage goes to 0
- I've had the hang occur for normal testing (not doing code coverage)
- when the hang on the normal test occurred, I was able to ptrace and it was blocked on a futex

So, I don't think it's a travis problem or code coverage problem or timeout problem. I've added a commit to disable self tests for now. I don't think kcov was a problem, but I'd still like to switch to tarpaulin.

philipc:
tarpaulin's coverage results don't make much sense, it seems a bit buggy. Still looking into it.

jonhoo:
@philipc yeah, we've seen coverage drop with tarpaulin over at https://github.com/mattnenterprise/rust-imap/pull/46 too. But I agree we probably want to change.

philipc:
tarpaulin's coverage has been improved, I'll try this again when there's a new release of it.

PR #56, opened 2017-10-04
Ignore self tests for now

philipc:
Experiencing frequent hangs when running them under code coverage tools.
I've also seen a hang running standalone too.

When the hang occurs, CPU usage drops to 0 and strace says the test
process is waiting on a futex.

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -36,11 +36,13 @@ fn release_fixture_path() -> (PathBuf, PathBuf) {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


jonhoo:
I guess ignoring them is the right thing to do for now, but I'd argue that it's at best a short-term solution.

philipc:
> I guess ignoring them is the right thing to do for now, but I'd argue that it's at best a short-term solution.

Definitely, there's a bug there that needs tracking down, but I don't really want to get sidetracked debugging it right now.

PR #59, opened 2017-10-11
Performance improvements

philipc:
See commits for details.

The second commit has a regression for the `build_default` benchmark. This appears to be due to the longer lifetimes of abbreviations. perf shows more time spent in jemalloc. It may be possible to avoid this if we had something like https://github.com/gimli-rs/gimli/issues/251.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -667,14 +667,45 @@ where
         &self,
         entry: &gimli::DebuggingInformationEntry<'a, 'b, gimli::EndianBuf<'input, Endian>>,
     ) -> Result<Option<Program<'input, Endian>>> {
-        let mut name = None;
+        let program = self.parse_program_attributes(entry).chain_err(|| {
+            format!(
+                "failed to parse attributes for subroutine at <{:x}><{:x}>",
+                self.header.offset().0,
+                entry.offset().0
+            )
+        })?;
+
+        let name = match program.name {
+            Some(name) => name,
+            None => return Ok(None),
+        };
+
+        // Where does this function live?
+        let ranges = if let Some(offset) = program.ranges {
+            self.parse_noncontiguous_ranges(offset)?
+        } else {
+            Self::parse_contiguous_range(program.low_pc, program.high_pc, program.size)?
+                .into_iter()
+                .collect()
+        };
+        if ranges.is_empty() {
+            return Ok(None);
+        }
+
+        Ok(Some(Program {
+            ranges,
+            inlined: entry.tag() == gimli::DW_TAG_inlined_subroutine,
+            name,
+        }))
+    }
+
+    fn parse_program_attributes<'a, 'b>(
+        &self,
+        entry: &gimli::DebuggingInformationEntry<'a, 'b, gimli::EndianBuf<'input, Endian>>,
+    ) -> Result<ProgramAttributes<'input, Endian>> {
+        let mut program = ProgramAttributes::default();
         let mut abstract_origin = None;
         let mut specification = None;
-        let mut low_pc = None;
-        let mut high_pc = None;
-        let mut size = None;
-        let mut range_offset = None;
-
         let mut attrs = entry.attrs();
         while let Some(attr) = attrs
             .next()
@@ -684,12 +715,11 @@ where
             match attr.name() {
                 // For naming, we prefer the linked name, if available
                 gimli::DW_AT_linkage_name | gimli::DW_AT_MIPS_linkage_name => {
-                    name = attr.string_value(&self.debug_str);
+                    program.name = attr.string_value(&self.debug_str);
                 }
-                gimli::DW_AT_name => if name.is_none() {
-                    // Linked name is not available, so fall back to just plain old name,
-                    // if that's available.
-                    name = attr.string_value(&self.debug_str);
+                gimli::DW_AT_name => if program.name.is_none() {
+                    // Linked name is not set yet, so fall back to just plain old name.
+                    program.name = attr.string_value(&self.debug_str);
                 },
                 gimli::DW_AT_abstract_origin => {
                     abstract_origin = Some(attr.value());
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -538,79 +556,94 @@ where
...
-                        .chain_err(|| "invalid low_pc attribute")
-                }
-                _ => match entry.attr_value(gimli::DW_AT_entry_pc) {
-                    Ok(Some(gimli::AttributeValue::Addr(addr))) => addr,
-                    Err(e) => {
-                        return Err(Error::from(ErrorKind::Gimli(e)))
-                            .chain_err(|| "invalid entry_pc attribute")
+            let mut attrs = entry.attrs();
+            while let Some(attr) = attrs
+                .next()
```
jonhoo: Is it correct to replace `next_dfs` with `next` here?

src/lib.rs
```diff
@@ -538,79 +556,94 @@ where
...
-                    return Err(Error::from(ErrorKind::Gimli(e)))
-                        .chain_err(|| "invalid low_pc attribute")
-                }
-                _ => match entry.attr_value(gimli::DW_AT_entry_pc) {
-                    Ok(Some(gimli::AttributeValue::Addr(addr))) => addr,
-                    Err(e) => {
-                        return Err(Error::from(ErrorKind::Gimli(e)))
-                            .chain_err(|| "invalid entry_pc attribute")
+            let mut attrs = entry.attrs();
+            while let Some(attr) = attrs
```
jonhoo: Hmm.. This now walks all the entries in the abbreviation rather than just the first in DFS order, which means that if an attribute is present in more than one, it'll override the others. Is that the right thing to do? Do we expect there to be many uninteresting entries? If so, we're now iterating over all of them whereas before we didn't do any iteration.

src/lib.rs
```diff
@@ -538,79 +556,94 @@ where
...
+                .next()
+                .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+                .chain_err(|| "invalid unit attribute")?
+            {
+                match attr.name() {
```
jonhoo: This change means that we no longer error out if we find that a compilation unit is missing a unit statement list, compilation unit name, compilation unit directory, etc. That seems unfortunate?

src/lib.rs
```diff
@@ -621,182 +654,203 @@ where
...
+        let mut size = None;
+        let mut range_offset = None;
+
+        let mut attrs = entry.attrs();
+        while let Some(attr) = attrs
+            .next()
+            .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+            .chain_err(|| "invalid subprogram attribute")?
+        {
+            match attr.name() {
```
jonhoo: We do essentially this match for the compilation unit too. Maybe it should be factored out?

src/lib.rs
```diff
@@ -621,182 +654,203 @@ where
...
-        if let Some(range) = Self::parse_contiguous_range(entry)?.map(|range| vec![range]) {
-            return Ok(range);
+            match attr.name() {
+                // For naming, we prefer the linked name, if available
+                gimli::DW_AT_linkage_name | gimli::DW_AT_MIPS_linkage_name => {
+                    if let Some(name) = attr.string_value(&self.debug_str) {
+                        return Ok(Some(name));
+                    }
+                }
+                gimli::DW_AT_name => {
```
jonhoo: this is missing `if name.is_none()`

src/lib.rs
```diff
@@ -621,182 +654,203 @@ where
...
+        while let Some(attr) = attrs
+            .next()
+            .map_err(|e| Error::from(ErrorKind::Gimli(e)))
+            .chain_err(|| "invalid subprogram attribute")?
         {
-            return Ok(range);
-        }
-        if let Some(range) = Self::parse_contiguous_range(entry)?.map(|range| vec![range]) {
-            return Ok(range);
+            match attr.name() {
```
jonhoo: This is essentially the same name matching as in `resolve_name`. Should be factored out.

src/lib.rs
jonhoo: ah, I guess you return early in the arm above, so it doesn't really matter..
should the function return if `name.is_some()` instead of calling `self.resolve_name` recursively below?

jonhoo:
So if I've understood this correctly, c730ccffc2b2e30426f940e1f3f204793893e036 is such a win because we only parse the units that we do lookups in? If we did enough lookups, then in theory we would still end up parsing every unit, and execution time would be the same as in the old setup?

src/lib.rs
philipc: There shouldn't be duplicate attributes. I don't think it matters if the behaviour has changed for this case.

Before we iterated multiple times up to a specific attribute. Now we iterate once over all attributes. There aren't many uninteresting attributes, so this is a win.

src/lib.rs
philipc: This behaviour hasn't been changed; we already didn't return an error if they were missing. The only change is that if the attribute does exist and we failed to parse it, then the error returned is now less specific.

src/lib.rs
philipc: It's hard to factor out because here there's some extra attributes that are needed (`low_pc`/`high_pc`/`ranges`).

src/lib.rs
philipc: Maybe. I'll see if I can make this a bit cleaner.

src/lib.rs
philipc: See reply to comment in `resolve_name`.

philipc:
> So if I've understood this correctly, c730ccf is such a win because we only parse the units that we do lookups in? If we did enough lookups, then in theory we would still end up parsing every unit, and execution time would be the same as in the old setup?

Exactly. This is the same with the way that we currently lazily parse `.debug_line`. It's not going to help if you eventually lookup an address in every compilation unit. But if you only do a few lookups (eg one backtrace) then it should be a win.

src/lib.rs
philipc: Actually, the error wasn't all the specific before anyway, because the error could have occurred on an attribute leading up to the one we were looking for.

philipc:
I've added a commit to cleanup the attribute and name parsing.

PR #60, opened 2017-10-12
Use cargo-tarpaulin instead of kcov

philipc:


.travis.yml
```diff
@@ -17,28 +17,23 @@ rust:
...
 after_success:
-- if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_RUST_VERSION" == "nightly" ]]; then travis-cargo coveralls --no-sudo --verify; fi
-- if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_RUST_VERSION" == "nightly" ]]; then KCOV=./kcov/build/src/kcov ./coverage; fi
-
-env:
-  global:
-  - TRAVIS_CARGO_NIGHTLY_FEATURE=nightly
+- if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_RUST_VERSION" == "stable" ]]; then
+    bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh);
+    cargo tarpaulin --no-count --ciserver travis-ci --coveralls $TRAVIS_JOB_ID;
```
jonhoo: Why `--no-count`?

philipc: Because it's slow and doesn't work properly. Note that kcov didn't do counts either.

It is slow because we're doing coverage on an integration test that loops over many test addresses. Avoiding the ptrace on subsequent loops speeds things up a lot.

It doesn't work properly because currently tarpaulin can't handle multithreaded code, and disables most of the counts anyway. rust's tests run in separate threads. There's an issue in tarpaulin to fix this.

Issue #11, opened 2016-12-05
Expose inlining chains, not just last step


jonhoo:
`Mapping::locate()` currently exposes only the *closest* function to a given address, which means that any address will only be resolved to the deepest inlined function that contains it. It would be better if `locate` returned the full chain of function ancestors for an address.

cuviper:
This would be like binutils `addr2line -i` right?  And `eu-addr2line` has this too.

jonhoo:
@cuviper that's exactly right.

fitzgen:
Now that `backtrace-rs` is using this crate (behind a feature flag) it seems even more important to fix this, as we backtraces will be incomplete without it.

Filed https://github.com/alexcrichton/backtrace-rs/issues/67 on the `backtrace-rs` side.

jonhoo:
I think we may want to consider making `Mapping::locate` return a `struct` instead of an increasingly long tuple along with this change. Otherwise, how would anyone remember that `.3` of the returned value is the call chain?

fitzgen:
Indeed. Ideally, we would return an iterator that didn't do work unless it was asked for too.

philipc:
Fixed by #64

PR #61, opened 2017-10-21
Add option to use symbol table

### LINKED ISSUE #3: addr2line should use .symtab information
In the absence of `.debug_line` information, there is still other file and address information in `.debug_info` that we can use to give vague information about the address.

philipc:
Fixes #3 

This gives results that match binutil addr2line for all addresses except for a couple of oddities (eg zero size symbol).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,9 @@ extern crate error_chain;
 
 use owning_ref::OwningHandle;
 use fallible_iterator::FallibleIterator;
+use object::SymbolKind;
 
+use std::cmp;
 use std::fmt;
 use std::path;
 use std::error;
@@ -143,6 +145,7 @@ pub use errors::*;
 #[derive(Clone, Copy, Default)]
 pub struct Options {
     with_functions: bool,
+    with_symbol_table: bool,
     with_demangling: bool,
 }
 
@@ -152,6 +155,14 @@ impl Options {
     /// address appears in. This comes at some parsing and lookup cost.
     pub fn with_functions(mut self) -> Self {
         self.with_functions = true;
+        self
+    }
+
+    /// Make the `Mapping` fallback to using the symbol table if there
+    /// is no debug information for an address. This comes at some parsing
+    /// and lookup cost.
+    pub fn with_symbol_table(mut self) -> Self {
+        self.with_symbol_table = true;
         self
     }
 
@@ -215,6 +226,7 @@ where
     Endian: gimli::Endianity,
 {
     units: Vec<Unit<'input, Endian>>,
+    symbols: Vec<Symbol<'input>>,
     opts: Options,
 }
 
@@ -250,7 +262,7 @@ impl Mapping {
     pub fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<str>>)>> {
+    ) -> Result<Option<(Option<path::PathBuf>, Option<u64>, Option<Cow<str>>)>> {
         self.inner.locate(addr)
     }
 }
@@ -272,7 +284,7 @@ impl<'input> BufferMapping<'input> {
     pub fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<'input, str>>)>> {
+    ) -> Result<Option<(Option<path::PathBuf>, Option<u64>, Option<Cow<'input, str>>)>> {
         self.0.locate(addr)
     }
 }
@@ -294,7 +306,7 @@ impl<'input> EndianDebugInfo<'input> {
     fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<'input, str>>)>> {
+    ) -> Result<Option<(Option<path::PathBuf>, Option<u64>, Option<Cow<'input, str>>)>> {
         match *self {
             EndianDebugInfo::LEInfo(ref mut dbg) => dbg.locate(addr),
             EndianDebugInfo::BEInfo(ref mut dbg) => dbg.locate(addr),
@@ -342,16 +354,23 @@ where
             }
         }
 
+        let symbols = if opts.with_symbol_table {
+            parse_symbols(file)
+        } else {
+            Vec::new()
+        };
+
         Ok(DebugInfo {
-            units: units,
-            opts: opts,
+            units,
+            symbols,
+            opts,
         })
     }
 
     pub fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<'input, str>>)>> {
+    ) -> Result<Option<(Option<path::PathBuf>, Option<u64>, Option<Cow<'input, str>>)>> {
         // First, find the compilation unit for the given address
         for unit in &mut self.units {
             if !unit.maybe_contains_address(addr) {
@@ -390,7 +409,7 @@ where
 
             unit.read_programs()?;
             if unit.programs.is_empty() {
-                return Ok(Some((path, line, None)));
+                return Ok(Some((Some(path), line, None)));
             }
 
             // The unit also has programs, so let's look for the function wrapping this address.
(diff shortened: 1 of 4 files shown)


philipc:
Failure is due to tarpaulin. I'm looking into it, but seems unrelated to this PR.

src/lib.rs
```diff
@@ -462,41 +481,120 @@ where
...
-    String::from_utf8_lossy(mangled)
+fn demangle_rust_symbol(mangled: &[u8]) -> Option<Cow<str>> {
+    None
+}
+
+struct Symbol<'input> {
+    name: &'input [u8],
+    file: Option<&'input [u8]>,
+    begin: u64,
+    end: u64,
```
fitzgen: Thinking out loud...

Demangling can be expensive -- we might want to add a refcell of the demangled symbol name here.

Ideally we would want this on some part of the common path for both debug info symbols and symbol table symbols.

But also probably not for very single symbol all the time... maybe an associative cache?

Maybe we just don't symbolicate addresses from the same function very often and this isn't worth it at all?

src/lib.rs
```diff
@@ -462,41 +481,120 @@ where
...
+                if symbol.name().len() == 0 {
+                    filename = None;
+                } else {
+                    filename = Some(symbol.name());
+                }
+                continue;
+            }
+            SymbolKind::Section | SymbolKind::Common | SymbolKind::Tls => continue,
+        }
+        if symbol.is_undefined() || symbol.name().len() == 0 || symbol.size() == 0 {
```
fitzgen: Here and just above: `symbol.name().is_empty()` is slightly more idiomatic than `symbol.name().len() == 0`.

src/lib.rs
```diff
@@ -462,41 +481,120 @@ where
...
+            end: symbol.address() + symbol.size(),
+        });
+    }
+
+    symbols.sort_by(|a, b| {
+        let ord = a.begin.cmp(&b.begin);
+        if ord != cmp::Ordering::Equal {
+            return ord;
+        }
+        a.end.cmp(&b.end)
```
fitzgen: This can be rewritten as

```rust
a.begin.cmp(&b.begin).then_with(|| a.end.cmp(&b.end)
```

src/lib.rs
```diff
@@ -462,41 +481,120 @@ where
...
+    });
+
+    let mut prev_end = 0;
+    for symbol in &mut symbols {
+        if symbol.begin < prev_end {
```
fitzgen: This fixup stuff deserves a comment.

src/lib.rs
jonhoo: Oooh, I didn't know that trick!

src/lib.rs
```diff
@@ -462,41 +481,120 @@ where
...
+        }
+        symbols.push(Symbol {
+            name: symbol.name(),
+            file: filename,
+            begin: symbol.address(),
+            end: symbol.address() + symbol.size(),
+        });
+    }
+
+    symbols.sort_by(|a, b| {
```
jonhoo: Would it perhaps be an idea to insert the symbols into a `BTreeMap` instead? I believe that should be faster.

src/lib.rs
```diff
@@ -250,7 +262,7 @@ impl Mapping {
     pub fn locate(
         &mut self,
         addr: u64,
-    ) -> Result<Option<(path::PathBuf, Option<u64>, Option<Cow<str>>)>> {
+    ) -> Result<Option<(Option<path::PathBuf>, Option<u64>, Option<Cow<str>>)>> {
```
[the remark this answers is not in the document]
philipc: Yeah, you're probably right. I left the `Option` there so that the caller can quickly tell if there was any info at all. We still need to convert this tuple to a struct, so can investigate getting rid of the `Option` as part of that.

src/lib.rs
philipc: Or maybe demangling should be deferred to the caller? We could return the language attribute so the caller knows how to demangle it.

src/lib.rs
philipc: Ah yeah, I often forget that in rust.

src/lib.rs
philipc: I don't think we want to use `BTreeMap`. I think what we want is a different array layout (see https://arxiv.org/abs/1509.05053). I'm not aware of an implementation of this in rust though.

jonhoo: I guess the question here is what we're expecting `n` to be, and the expected number of searches. Figure 17 of that paper is likely the comparison we want. When `n` is large (as in > =2^18, or perhaps smaller depending on the efficiency of Rust's binary search implementation), or when the number of lookups is small, I believe a `BTreeMap` is the appropriate structure until someone does implement Eytzinger (the former is supported by the paper, the latter by the fact that the array also needs to be sorted in order to do binary searches). I suspect that we expect `n` to be fairly small, though I don't have a good sense of the number of symbols in a representative symbol file. It'd be useful if the paper above also included the cost of sorting, but a cursory reading suggests that they don't.

Without an implementation of Eytzinger (which seems like it might be interesting), I think `BTreeMap` is still the right way to go. For small `n`, latency will be pretty short anyway, and for large `n`, `BTreeMap` wins.

jonhoo: Reading a bit more closely, their B-tree isn't anywhere near the same as a Rust `BTreeMap`. It assume contiguous storage. Hard to say exactly how the two would compare, but Rust's would likely be slower. I guess then I'm fine with the current code, even though I feel like the sorting will actually introduce a non-trivial up-front cost that might not be right trade-off if we expect few lookups.

jonhoo: @philipc fyi: https://github.com/patmorin/arraylayout/issues/3

philipc: Yes, I meant we want to use contiguous storage as currently, but with a better layout and search algorithm than binary search, so not rust's `BTreeMap` because it is not contiguous. Whether that sorting strategy should be B-tree or Eytzinger needs testing.

> I guess then I'm fine with the current code, even though I feel like the sorting will actually introduce a non-trivial up-front cost that might not be right trade-off if we expect few lookups.

So I'm not familiar with the costs of `BTreeMap` construction, but I didn't expect it to be significantly better than a sort?

philipc: I should also point out that the `object` crate is already sorting Mach-O symbols internally to calculate symbol sizes (but it doesn't give us the sorted array), and there's a bunch of inefficiencies due to creating multiple arrays of these symbols (eg for ELF there's one in the `goblin` crate, one in the `object` crate and one here) so that needs optimizing too. Ideally `goblin` and `object` would both give iterators instead, but I don't know how to get that to work with the Mach-O symbol sizes.

jonhoo: > So I'm not familiar with the costs of `BTreeMap` construction, but I didn't expect it to be significantly better than a sort?
I believe it should be faster, as you avoid walking all the elements once at the beginning to push them onto the list. The sorting itself is probably roughly the same, although that depends on the exact algorithm Rust uses internally (since it might shift things in the array). You might also want to use `sort_unstable_by` here.

Ah, I wasn't aware of the sorting in `object`. I don't know which of these approaches are best suited for adopting an already-sorted array, though it might very well be `push` + `sort_by`.

jonhoo: @philipc First implementation here: https://github.com/jonhoo/ordsearch

src/lib.rs
philipc: I tried to write a comment to justify this, but it made no difference on any of my test binaries, so I've deleted it.

src/lib.rs
philipc: Nice.  I'm going to leave optimisation of this for future work.

philipc:
Tarpaulin failure should be fixed soon. Nightly failure is unrelated to this PR, I'll investigate separately.

src/lib.rs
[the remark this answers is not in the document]
main--: fwiw, the roadblock I hit when trying to work with gimli-addr2line is the lack of inlining info. Ultimately I think this needs to return an iterator, like I'm doing here: https://github.com/main--/unwind-rs/blob/master/hardliner/src/lib.rs#L368

src/lib.rs
main--: >Or maybe demangling should be deferred to the caller?

This is what I decided to do:

https://github.com/main--/unwind-rs/blob/988932544a01621109d476cb7a2381d71d886fb8/hardliner/src/lib.rs#L246

(other than the lack of inlining info, I wrote my hardliner-addr2line implementation also because gimli-addr2line's API design feels a little bolted-on (pass options in, get data out is weird unless you're a command line tool that wants to do exactly this))

src/lib.rs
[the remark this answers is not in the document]
philipc: Yep the current plan in #11 is to return an iterator.

src/lib.rs
philipc: I agree about the API. I'm not sure that the memmap belongs in this crate.

sfackler:
This appears to have been undone by #64 - was that intentional?

fitzgen:
IIUC, the intention is to have `object` provide the symbol table interface. It would be nice if there was an EZ mode for users who just want file/function/line/column/etc and don't care about which source the information came from...

cc @philipc @main--

sfackler:
I kind of imagine addr2line as the place for that EZ mode interface, right? "Give me as much information as possible about this address."

main--:
What you're describing is *one* possible usecase but that's not the right abstraction for everyone. As #64 introduced inlining support, addr2line now always gives you complete traces without any missing frames. We can't uphold this guarantee if we silently fall back to symtab however. Thus, symbol lookup would need to be its own separate interface.

The old API was basically a straight adaptation from the `addr2line` console application, but the new one aims to be more flexible. One part of that (not done yet) is that we want to make the `object` dependency optional so you can use your own parser (if necessary). But this means that you can only use `addr2line` in one of two ways:

* Either you use it with `object` (the only option right now) and you construct the object on your side, then pass a reference to us. In that case you're already holding the object and you can just get the symbols from there.
* Or you provide a custom implementation and we would literally just pass take symbols from you and pass them back out.

So the consensus in #63 was to keep symbol table handling in the `object` crate.

sfackler:
Cool - it makes sense that you'd want a library to handle the DWARF-side of symbolication specifically. Might make sense to rename it at some point though, since `addr2line` (to me at least) implies a more general kind of thing like what I described earlier.

philipc:
I'm fine with adding an optional EZ mode if it's useful. But as soon as you start needing to configure the behaviour then you're better off just using the flexible APIs. So the addr2line example wouldn't use any EZ mode we added.

PR #64, opened 2017-11-07
Merge hardliner

### LINKED ISSUE #63: API flexibility
Prompted by @main--'s [comment](https://github.com/gimli-rs/addr2line/pull/61#discussion_r146528613), here's a few API changes for consideration. The common theme is to make the API more flexible, and less centered on providing an addr2line clone. Note that more flexible may also mean more complicated, so this is a tradeoff. Some of these may be strawman proposals :)

- Remove the `with_functions` option, and use a parameter to `locate` instead (or a different `locate` method). This allows the caller to decide when doing the `locate`, not when constructing the `Mapping`. This is possible because we lazily read the function information.

- Remove the `with_demangling` option and stop doing demangling. This should be handled by the caller. Instead, we return a language value that the caller can use to determine how to demangle. This allows the caller to access both the mangled and demangled names if desired, and to implement their own caching of demangled names if required.

- Remove the `with_symbol_table` option. The caller should be able to decide whether to use symbols for each query. The caller should also be able to get both debuginfo and symbol table information simultaneously. This may mean having two independent queries: one for debuginfo and one for the symbol table.

- Make the `object` crate dependency optional. Instead, define a trait for the required object loader functionality. We can still implement this trait for `object` for convenience.

- Avoid the `memmap` dependency. Using an `OwningHandle` to tie something to a memmap is a generic operation that has nothing to do with addr2line, and should be implemented elsewhere.

- Add an endianity parameter to the mapping instead of using `EndianDebugInfo`. This allows the caller to decide if they only want to support `NativeEndian`, without incurring size or performance penalties.

main--:
Missing pieces (that I can think of):

- [ ] I *think* I fixed `benchmark.sh` but I didn't test that.
- [ ] I didn't touch the cargo benchmark as I'm working on stable Rust.
- [ ] README, docs

Fixes #63.


philipc:
I think this is fine for the purposes of bringing in hardliner. The missing stuff can be addressed in followup PRs.

We'll probably want to move the addr2line example back to bin. I don't think there's any downside to doing that?

main--:
>We'll probably want to move the addr2line example back to bin. I don't think there's any downside to doing that?

Doesn't that force a `clap` dependency even for library users? I really wouldn't like that.

philipc:
Ah yep, cargo doesn't yet support dependencies that only apply to bin, so ignore that.

philipc:
@main-- You should have permission to merge this if you think it is ready. I was unsure what your plan with the checkbox items was.

PR #66, opened 2017-11-10
Handle more name attributes

philipc:
Also some addr2line output changes to make it easier to compare against other tools.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -424,30 +424,50 @@ impl<R: gimli::Reader> FullContext<R> {
 
 type Error = gimli::Error;
 
-fn str_attr<'abbrev, 'unit, R: gimli::Reader>(
+fn name_attr<'abbrev, 'unit, R: gimli::Reader>(
     entry: &gimli::DebuggingInformationEntry<'abbrev, 'unit, R, R::Offset>,
     dw_unit: &gimli::CompilationUnitHeader<R, R::Offset>,
     abbrevs: &gimli::Abbreviations,
     sections: &DebugSections<R>,
-    name: gimli::DwAt,
 ) -> Result<Option<R>, Error> {
-    Ok(match entry.attr(name)? {
-        Some(x) => x.string_value(&sections.debug_str),
-        None => {
-            match entry.attr_value(gimli::DW_AT_abstract_origin)? {
-                Some(gimli::AttributeValue::UnitRef(offset)) => {
-                    let mut tcursor = dw_unit.entries_at_offset(abbrevs, offset)?;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -424,30 +424,50 @@ impl<R: gimli::Reader> FullContext<R> {
...
+    if let Some(attr) = entry.attr(gimli::DW_AT_MIPS_linkage_name)? {
+        if let Some(val) = attr.string_value(&sections.debug_str) {
+            return Ok(Some(val));
+        }
+    }
+    if let Some(attr) = entry.attr(gimli::DW_AT_name)? {
+        if let Some(val) = attr.string_value(&sections.debug_str) {
+            return Ok(Some(val));
+        }
+    }
```
main--: I don't like the code duplication here. Perhaps something like `[gimli::DW_AT_linkage_name, gimli::DW_AT_MIPS_linkage_name, gimli::DW_AT_name].iter().filter_map(|x| entry.attr(x)).filter_map(|x| x.string_value(&sections.debug_str)).next()` helps?

src/lib.rs
```diff
@@ -424,30 +424,50 @@ impl<R: gimli::Reader> FullContext<R> {
...
+    match entry.attr_value(gimli::DW_AT_abstract_origin)? {
+        Some(gimli::AttributeValue::UnitRef(offset)) => {
+            let mut entries = dw_unit.entries_at_offset(abbrevs, offset)?;
+            if let Some((_, entry)) = entries.next_dfs()? {
+                // FIXME: evil dwarf can send us into an infinite loop here
```
main--: Does the spec even allow chains of redirections? If not, do compilers adhere to that? Then we could solve this with a simple flag.

src/lib.rs
philipc: `DW_AT_linkage_name` and `DW_AT_MIPS_linkage_name` will be combined when we do something like [this commit](https://github.com/gimli-rs/addr2line/pull/59/commits/25471236e22d7a33a6307df915eee73ae4f003b7). Might be hard to completely avoid duplication for `DW_AT_name` though.

src/lib.rs
philipc: Technically I think it's not allowed, but gdb has [a comment](https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=gdb/dwarf2read.c;h=b6bd2917e35f378e983cf959d1273399ab8ccf65;hb=HEAD#l12326) saying that gcc generates it. I think a simple recursion count with a low limit will be enough.

PR #68, opened 2017-11-10
Fix code coverage

philipc:
Stop tarpaulin passing a --quiet flag to the tests, which fails with:
```
thread 'main' panicked at '"Unrecognized option: \'quiet\'"', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-test-0.2.0/src/lib.rs:266:26
```
Will investigate further later.


philipc:
Runs the tests, but now they hang, so that's no good either.

PR #71, opened 2017-11-26
Fix #51

### LINKED ISSUE #51: Handle references to other compilation units
For `DW_AT_abstract_origin`/`DW_AT_specification`, we handle `UnitRef` values, but currently don't handle `DebugInfoRef`.

main--:


src/lib.rs
```diff
@@ -465,28 +472,27 @@ fn name_attr<'abbrev, 'unit, R: gimli::Reader>(
...
-    }
-    match entry.attr_value(gimli::DW_AT_specification)? {
-        Some(gimli::AttributeValue::UnitRef(offset)) => {
-            let mut entries = dw_unit.entries_at_offset(abbrevs, offset)?;
-            if let Some((_, entry)) = entries.next_dfs()? {
-                // FIXME: evil dwarf can send us into an infinite loop here
-                return name_attr(entry, dw_unit, abbrevs, sections);
-            }
+        Some(gimli::AttributeValue::DebugInfoRef(dr)) => {
+            if let Some((unit, offset)) = sections.debug_info.units()
```
philipc: We've already got a Vec of parsed compilation unit headers that we could use?

src/lib.rs
```diff
@@ -465,28 +472,27 @@ fn name_attr<'abbrev, 'unit, R: gimli::Reader>(
...
+                .filter_map(|unit| dr.to_unit_offset(&unit).map(|uo| (unit, uo))).next()? {
+                    let mut entries = unit.entries_at_offset(abbrevs, offset)?;
```
philipc: Also these abbreviations will be for the wrong unit.

PR #73, opened 2017-12-14
 Add symbol table fallback to addr2line example 

philipc:
It would have been nice to change `FunctionName` to hold `Cow<str>` instead of `gimli::Reader` so that symbols could be `FunctionName`s too, but the lifetime on `Reader::to_string_lossy` doesn't allow that.

Demangling is a bunch of small functions again, but I don't see how to do better.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "gimli-rs/addr2line" }
 [dependencies]
 gimli = "0.14"
 fallible-iterator = "0.1"
-object = "0.4"
+object = "0.7"
 intervaltree = "0.2"
 smallvec = "0.4"
 rustc-demangle = { version = "0.1", optional = true }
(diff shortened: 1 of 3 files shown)


philipc:
Note the following about GNU addr2line "auto" demangling:
- It's very C++ centric, "auto" mostly means to guess which C++ demangling scheme to use.
- It does also guess when to use Rust demangling, but not Java or D.
- It doesn't use the `DW_AT_language` value.
- It does demangle symbol table names with demangling is set to "auto".

So I don't think we should follow GNU addr2line too closely. A better behaviour for "auto" would be:
- use `DW_AT_language` if known
- otherwise inspect the symbol to determine if it's a known mangling scheme. Trying C++ demangling follow by rust demangling is a first attempt at this, but we could certainly be smarter.

My reason for changing `demangle`'s return type was to avoid an extra `to_string_lossy`, but I don't feel strongly about that.

main--:
Okay, makes sense. I still think this should be handled on the CLI side though - guesswork and fallbacks don't feel like a clean and well-designed API to me.

philipc:
Moved the demangle fallback into the CLI. demangle probably doesn't belong in this library at all after this, but I didn't change that.

PR #75, opened 2017-12-20
Lazy parsing

philipc:


main--:
Ah, I guess the problem is that that would keep a `&mut` borrow on `self` alive?

In that case a (ref)cell may be worth considering - it has the added bonus of not requiring `&mut` for the query functions either.

philipc:
Yeah it was due to the `&mut` borrow, but I just did similar to what we used to do before hardliner, instead of thinking about it much; I'll look into it more tomorrow. `Cell` isn't appropriate because it needs to move the value in and out of the `Cell`. I've not had much experience with `RefCell`, but I think that would be trading `&mut` for runtime panics if you happen to use this from multiple threads, so I don't think that works either?

philipc:
There's also https://crates.io/crates/lazy-init

main--:
>I've not had much experience with `RefCell`, but I think that would be trading `&mut` for runtime panics if you happen to use this from multiple threads, so I don't think that works either?

No, what `RefCell` does is simply that it moves the borrow checker from compile time to runtime (at least that's how I think of it). It's basically the single-threaded version of `RwLock`.

I didn't consider multi-threaded environments. In that case, lazy-init does seem like a better solution.

philipc:
Added a commit for lazy-init.

I looked into using `RefCell` but it was starting to get complicated because the `Ref` needed to live as long as `IterFrames`.

src/lib.rs
```diff
@@ -197,66 +207,76 @@ impl<'a> Context<gimli::EndianBuf<'a, gimli::RunTimeEndian>> {
...
+                self.comp_name.clone(),
+            )?;
+
+            let (lnp, mut sequences) = ilnp.sequences()?;
+            sequences.retain(|x| x.start != 0);
+            sequences.sort_by_key(|x| x.start);
+            Ok(Lines { lnp, sequences })
+        }) {
+            Ok(ref lines) => Ok(lines),
+            Err(e) => Err(e),
```
main--: Took me a while to figure out what this does. Perhaps `.as_ref().map_err(Error::clone)` is clearer?

philipc: Yeah that looks better, thanks.

PR #78, opened 2018-01-05
deps: update memmap to 0.6

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -8,6 +8,7 @@ extern crate object;
 extern crate test;
 
 use std::env;
+use std::fs::File;
 use std::path::{self, PathBuf};
 use std::process;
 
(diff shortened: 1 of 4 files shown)


benches/bench.rs
```diff
@@ -22,8 +22,8 @@ fn release_fixture_path() -> PathBuf {
 }
 
 fn with_file<F: FnOnce(&object::File)>(target: &path::Path, f: F) {
```
philipc: Need to open the file here (like you did in the other conversions).

ignatenkobrain: Right, missed this part.

ignatenkobrain: Should be fixed now.

PR #80, opened 2018-01-06
Don't check symbols containing `{{impl}}` in correctness test

philipc:
nightly rust has fixed the linkage name for these, so they no longer
contain `{{impl}}`. Stop checking these so that we work on both stable
and nightly.


philipc:
I didn't test this correctly, and it won't work.

PR #92, opened 2018-06-02
Update object to 0.8.0 

est31:
Please take a close look before merging, as I don't think that we want to expose an implementation detail (arena).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "gimli-rs/addr2line" }
 [dependencies]
 gimli = "0.16"
 fallible-iterator = "0.1"
-object = "0.7"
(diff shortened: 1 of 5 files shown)


philipc:
This is the part that I had started work on already, rather than general `no_std` changes for `addr2line`. Yeah, rental could be used to put the arena in the context, but the way that I've been proceeding is to use an `Rc` or `Arc` version of `gimli::EndianReader`, which still has reasonable performance. I've just noticed that `gimli::EndianReader` doesn't work with `Arc` currently though (https://github.com/gimli-rs/gimli/pull/305).

est31:
Ah I see. Closing then.

est31:
(in favour of your changes)

PR #93, opened 2018-06-02
Update object to 0.9.0

philipc:
Also use the `Object` trait instead of `File` (see https://github.com/gimli-rs/object/issues/53).

~~This requires https://github.com/gimli-rs/gimli/pull/305.~~

I played a bit with using `Rc/Arc<Cow<'input, [u8]>>` instead of `EndianRcSlice`, but wasn't sure it was worth the API complexity. So I think `Context::new` should be the simple high level interface, and consumers with more specific needs can use `Context::from_sections`.

Before:
```
test context_new_and_query_location       ... bench:     951,031 ns/iter (+/- 79,277)
test context_new_and_query_with_functions ... bench:   1,760,225 ns/iter (+/- 54,934)
test context_new_location                 ... bench:     760,223 ns/iter (+/- 57,231)
test context_new_with_functions           ... bench:     751,183 ns/iter (+/- 55,055)
test context_query_location               ... bench:  10,378,538 ns/iter (+/- 92,496)
test context_query_with_functions         ... bench:  11,578,429 ns/iter (+/- 142,557)
```

After:
```
test context_new_and_query_location       ... bench:   5,527,469 ns/iter (+/- 285,657)
test context_new_and_query_with_functions ... bench:   6,809,893 ns/iter (+/- 490,899)
test context_new_location                 ... bench:   5,230,294 ns/iter (+/- 307,209)
test context_new_with_functions           ... bench:   5,213,000 ns/iter (+/- 266,617)
test context_query_location               ... bench:  15,353,989 ns/iter (+/- 309,998)
test context_query_with_functions         ... bench:  16,591,755 ns/iter (+/- 402,449)
```

For reference, `Arc<Cow<[u8]>>`:
```
test context_new_and_query_location       ... bench:   1,063,159 ns/iter (+/- 57,010)
test context_new_and_query_with_functions ... bench:   2,483,267 ns/iter (+/- 180,353)
test context_new_location                 ... bench:     758,812 ns/iter (+/- 16,336)
test context_new_with_functions           ... bench:     755,755 ns/iter (+/- 31,252)
test context_query_location               ... bench:  18,362,107 ns/iter (+/- 417,995)
test context_query_with_functions         ... bench:  19,264,415 ns/iter (+/- 260,564)
```
The query numbers for `Arc<Cow<[u8]>>` are surprising to me, but I didn't look into it.

PR #94, opened 2018-06-07
Switch to lazycell

est31:
Lazycell has `#![no_std]` support. We give up sync-ness, but that's not needed for std inclusion.


philipc:
Can you also remove all the `R: Sync` bounds which were only there because `Lazy` needed them?

est31:
@philipc I've removed the Sync bounds. With this change, will #93 still depend on your gimli PR https://github.com/gimli-rs/gimli/pull/305 ?

philipc:
No it won't, I'll redo #93 after this is merged.

PR #97, opened 2018-07-05
Bump to 0.7.0

fitzgen:
+cc @alexcrichton

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "Apache-2.0/MIT"
 name = "addr2line"
 readme = "./README.md"
 repository = "https://github.com/gimli-rs/addr2line"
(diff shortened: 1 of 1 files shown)


philipc:
What's motivating this release? Note that the no_std work hasn't been finished off yet.

alexcrichton:
Oh sorry I don't mean to rush this at all, if it's not ready no worries! Right now the `backtrace` crate uses addr2line as an optional dependency, but addr2line as published on crates.io doesn't depend on the most recent version of gimli. (but the git version of addr2line does)

It's certainly not urgent at all, so if there's work to do feel free to disregard!

philipc:
Ok, I should be able to find time this weekend to finish off the no_std work, it'd be good to have that in the release.

PR #98, opened 2018-07-08
Implement no_std support

### LINKED ISSUE #90: #![no_std] support
In order for addr2line to be integrated into libstd, it requires support for the `#![no_std]` environment. This also helps with integration into embedded programs and OS kernels.

philipc:
Note that cpp_demangle and gimli may still need updating once they release updates, but they shouldn't be breaking changes.

Some of this is based on @est31's work.

Fixes #90


philipc:
Recent changes to the unstable alloc API have broken things (eg btree_map), and this requires fixes in other crates. Probably worth waiting to see what happens as a result of the alloc stabilization discussions.

PR #102, opened 2018-08-04
Implement no_std support

### LINKED ISSUE #90: #![no_std] support

philipc:
Fixes #90 

I also changed the travis builds so there's not quite so many builds.

This still needs a gimli update, so the nightly travis build will fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,33 +22,24 @@ rust:
 - stable
 
 env:
-- GIMLI_JOB="test"  GIMLI_PROFILE=
-- GIMLI_JOB="test"  GIMLI_PROFILE="--release"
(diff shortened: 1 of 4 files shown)


roblabla:
CC @main-- .

The only other thing unwind-rs' fork had was a bandaid that would parse `DW_CFA_GNU_args_size`, and ignore this result. From what I read, that implementation is wrong, and anyways is only necessary for i386 windows binaries (source: https://reviews.llvm.org/D38680)

I started a proper implementation for this. I guess we can wait for that before doing a gimli minor bump. I'll try to get a PR up during the week-end.

philipc:
This should be ready for review/merge now.

PR #107, opened 2018-11-27
Update to object 0.11

cuviper:


cuviper:
This completes #105 by fixing the use of `Object` lifetimes.  It might be a breaking change on `Context::new` to have the explicit lifetime on `&'file O` though, which was just `&O` before.

PR #111, opened 2019-02-06
Set default R for Context and release 0.8.0

philipc:


philipc:
I think all the failures are due to handling of backslashes when appending paths:
```
system addr2line: /rustc/4b1e39b7b36d677803e40130ea29ee6d300abf6e//src/libpanic_unwind/lib.rs:92
gimli addr2line:  /rustc/4b1e39b7b36d677803e40130ea29ee6d300abf6e/src/libpanic_unwind/lib.rs:92
```

philipc:
I'm fine with merging and releasing this. I think our behaviour is correct.

PR #115, opened 2019-02-21
Update gimli

philipc:
Includes more support for DWARF 5 string attribute forms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,21 +71,16 @@ struct Func<T> {
 }
 
 struct Lines<R: gimli::Reader> {
-    lnp: gimli::CompleteLineNumberProgram<R>,
-    sequences: Vec<gimli::LineNumberSequence<R>>,
+    lnp: gimli::CompleteLineProgram<R, R::Offset>,
+    sequences: Vec<gimli::LineSequence<R>>,
 }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -290,44 +281,41 @@ impl<R> ResUnit<R>
 where
     R: gimli::Reader,
 {
-    fn parse_lines(&self, sections: &DebugSections<R>) -> Result<&Lines<R>, Error> {
+    fn parse_lines(&self) -> Result<Option<&Lines<R>>, Error> {
+        let ilnp = match self.dw_unit.line_program {
+            Some(ref ilnp) => ilnp,
+            None => return Ok(None),
```
main--: This [used to be](https://github.com/gimli-rs/addr2line/pull/115/files#diff-b4aea3e418ccdb71239b96952d9cddb6L218) a straight `unreachable!()`, so I believe it should be extremely uncommon. From this perspective returning an error here *may* be warranted (if the line program for my unit is missing entirely, I probably want to know that instead of just being told "no idea where this address maps"). Not sure though.

philipc: I'm not sure either. Returning `None` lets us get any function information if available. But yes, it should be extremely uncommon.

PR #124, opened 2019-05-10
Fix compile error

gz:
The combination of default-features = false, features = ["alloc", "rustc-demangle"] fails to compile since the library is missing an import to format! in this case.


philipc:
Please make sure the other builds continue to work.

philipc:
This appears to be fixed, possibly by #156

PR #127, opened 2019-06-27
Fix build with `--no-default-features`

alexcrichton:
Avoid having a defaulted type parameter with `EndianRcSlice` when no
features are activated to ensure that the library builds correctly.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -21,6 +21,7 @@ case "$GIMLI_JOB" in
         ;;
 
     "nightly_features")
+        cargo build --no-default-features
         cargo build --no-default-features --features "alloc"
         cargo build --no-default-features --features "alloc object"
         ;;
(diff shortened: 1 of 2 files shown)


philipc:
What's the use case for building this without the alloc feature? I don't think it can do anything useful without allocating.

alexcrichton:
Ah I see, I was mostly just poking around at https://github.com/rust-lang/backtrace-rs/issues/189 and trying to slim the dependencies as much as possible (in number and enabled features). I was able to work with this crate with no features enabled, but I didn't realize that it still links to `alloc` and uses it even if the feature is disabled. I originally wanted to fix this to simply fix the build of `cargo build --no-default-features` (which is currently broken), but I see what you mean now!

Seems like this is probably not necessary in that case regardless!

PR #130, opened 2019-08-10
Improve performance for line number program handling

philipc:
Fixes some of https://github.com/rust-lang/backtrace-rs/issues/232#issuecomment-517578935.


philipc:
The lifetime on `Frame` and `Location` is a breaking change.

PR #133, opened 2019-09-04
Ignore broken units in Context<R>::from_sections.

### LINKED ISSUE #132: addr2line panics on XUL.dSYM from recent Firefox macOS builds
Steps to reproduce:
```
wget https://symbols.mozilla.org/XUL/3378E7F8C8E339A6AAAD847F4A12748D0/XUL.dSYM.tar.bz2
tar -xvf XUL.dSYM.tar.bz2
git clone https://github.com/gimli-rs/addr2line
cd addr2line
cargo run --example addr2line -- -i -f -e ../XUL.dSYM/Contents/Resources/DWARF/XUL 14c601b
```

Expected results:
A call stack

Actual results:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MissingUnitDie', src/libcore/result.rs:999:5
```

---

`Context<R>::from_sections` encounters `Error::MissingUnitDie` on 439 out of 23185 units. The broken units are spread out evenly.
In https://github.com/gimli-rs/gimli/issues/436, @philipc said:
> I think this is probably an addr2line bug. It should ignore units that it can't parse.

mstange:
Fixes #132. r? @philipc 

I don't know if I'm using the right words in the commit message, how to find out what the type of the "broken" units is or what this means about the validity of the DWARF data. I also don't know how to slim the testcase down into something that could become a real test.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,7 +174,10 @@ impl<R: gimli::Reader> Context<R> {
         let mut units = sections.units();
         while let Some(header) = units.next()? {
             let unit_id = res_units.len();
-            let dw_unit = sections.unit(header)?;
+            let dw_unit = match sections.unit(header) {
+                Ok(dw_unit) => dw_unit,
+                Err(_) => continue,
+            };
 
             let lang;
             {


philipc:
> how to find out what the type of the "broken" units is or what this means about the validity of the DWARF data

Do you mean how to find out the manner in which they are broken?

From the debug output you gave in the gimli issue, we can see that the length of the unit is 7 bytes. This is enough to contain the DWARF version number, the .debug_abbrev offset, and the address size. There are 0 bytes of entries (we could do a better job in the `Debug` implementation for `EndianReader` to make this clearer). This is unusual, and I can't see any point for it existing, but we can easily ignore it, so we should.

I think this PR is fine. Another option would be to only ignore units that match this specific instance of brokenness (`Error::MissingUnitDie`), but I don't know which approach is best.

For example, we could also easily fail parsing if the DWARF actually was corrupted, and if that happened do users prefer to find out the DWARF is corrupted, or do they prefer a best-effort answer? I think ideally the addr2line utility would print a warning about corruption and continue on, but the crate API doesn't allow that currently.

mstange:
> > how to find out what the type of the "broken" units is or what this means about the validity of the DWARF data
> 
> Do you mean how to find out the manner in which they are broken?

Yes, and whether it's "valid" brokenness, and if not, if this means that there's a bug in dsymutil or in llvm.

> From the debug output you gave in the gimli issue, we can see that the length of the unit is 7 bytes. This is enough to contain the DWARF version number, the .debug_abbrev offset, and the address size. There are 0 bytes of entries (we could do a better job in the `Debug` implementation for `EndianReader` to make this clearer). This is unusual, and I can't see any point for it existing, but we can easily ignore it, so we should.

I see, thanks!

> For example, we could also easily fail parsing if the DWARF actually was corrupted, and if that happened do users prefer to find out the DWARF is corrupted, or do they prefer a best-effort answer? I think ideally the addr2line utility would print a warning about corruption and continue on, but the crate API doesn't allow that currently.

From my very limited viewpoint, I'm mostly interested in a best-effort answer. I'd only care about corruption errors if there was some certainty that the corruption was in fact responsible for inaccurate or incomplete answers. Then those errors might enable me to find bugs in other parts of the pipeline, and fixing those bugs might give me more accurate and more complete answers.

PR #139, opened 2019-09-09
Use object crate for bench addresses

RReverser:
This should be both faster and more reliable across different distros.


philipc:
Thanks!

PR #141, opened 2019-09-10
Improve function lookup perfomance

philipc:
WIP. This is intended to address the remainder of https://github.com/rust-lang/backtrace-rs/issues/232#issuecomment-517578935.

It currently requires a new low level DIE parser in gimli, but I want to do more investigation to see if that can be avoided.


philipc:
This is all covered by other PRs now.

PR #147, opened 2019-11-23
chore: Upgrade object to 0.16, smallvec to 1.0

DanSnow:
Closed #144
Closed #145


DanSnow:
They didn't cause any problem.

Issue #149, opened 2019-11-28
Add location_iter api


eqv:
For some use-cases it would be super helpful to be able to iterate over all units/frames or locations in the binary. This could be accomplished simply by exposing the  `Lines` struct.

philipc:
Can you give an example of this use case?

Without knowing more, I think you may be better off using gimli directly. Exposing `Lines` is undesirable because the problem this crate solves is how to cache the data efficiently for address lookups, and we don't want to be locked into a particular algorithm (and indeed, `Lines` is something that was changed recently).

eqv:
I'm doing mass lookups (think: all instruction), in a linear manner to build a complete instruction => line mapping. This has two usecase: One to obtain a map of all lines that contributed to the binary, and two, to turn a large set of instruction level traces into source level traces. Doing a binary search for each address is terrible slow in this case.
By iterating directly over the ranges, I can build a hashmap Addr-> sourcfile/line, which cuts the computation time from nearly half an hour to seconds. 
 This crate already does a lot work to get all the gimli data into a format that I can use (sorted array, anything iteratable really). ATM I simply copied most of this crate into my code to access the internal representation, however this makes it much harder to keep up with bugfixes in addr2line and gimli.  If you think this usecase is probably uncommon, feel free to close this issue. 
Otherwise, I would probably submit a pull request for a API that exposes an iterator on `LineRow` or a similar data type, to limit exposing implementation details.

philipc:
How do you plan to handle the inline function information? I'm not sure that a hashmap is suitable for that.

If adding the iterator is relatively small, then I'm fine with merging that.

philipc:
Closing, see #153.

PR #153, opened 2019-12-07
added api that exposes the whole source mapping

eqv:
A proposal for #149

The rational behind this design:

- it avoids exposing any of your internal data structures
- it avoids adding any additional dependencies by returning a reference (HashMap could have been useful to create a owned version of all paths across the individual units). 

Would you feel comfortable to maintain this? If not, what needs to be changed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -370,6 +370,39 @@ impl<R: gimli::Reader> Context<R> {
         }
         Ok(())
     }
+
+    /// Returns a copy of the whole Source Map.
+    pub fn copy_line_infos(&self) -> Vec<LineInfo> {
+        let mut res = vec![];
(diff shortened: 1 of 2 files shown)


philipc:
I'm not comfortable maintaining this for two reasons:
- I feel this API is suboptimal. In particular, it is discarding the information about where each sequence ends.
- It is only loosely related to the rest of the addr2line crate and gains little from being in the same crate. I think you would be better off using gimli directly. If you want, I can put together an example of how to do it using gimli directly, and if that example turns out to be too long or complicated then I'm open to adding improvements to gimli to make that easier.

eqv:
Ok, the end of sequence issue could easily be fixed, but I see your other point. 
Thanks for considering it!

PR #159, opened 2020-04-30
Allow access to the parsed dwarf 

bjorn3:
This makes it possible get access to the internal parsed `Dwarf` and `Unit`. It also allows access to the `UnitOffset` corresponding to the DIE of a frame. This is very useful for https://github.com/bjorn3/pretty_backtrace/tree/var_guard.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -269,7 +269,12 @@ impl<R: gimli::Reader> Context<R> {
         })
     }
 
-    fn find_unit(&self, probe: u64) -> Option<usize> {
+    /// The dwarf sections associated with this `Context`.
+    pub fn dwarf(&self) -> &gimli::Dwarf<R> {
+        &self.sections
+    }
+
+    fn find_unit_id(&self, probe: u64) -> Option<usize> {
         let idx = self.unit_ranges.binary_search_by(|r| {
             if probe < r.0.begin {
                 Ordering::Greater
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -99,9 +99,12 @@ impl Context<gimli::EndianRcSlice<gimli::RunTimeEndian>> {
...
             Endian: gimli::Endianity,
         {
+            use object::ObjectSection;
+
             let data = file
-                .section_data_by_name(S::section_name())
-                .unwrap_or(Cow::Borrowed(&[]));
+                .section_by_name(S::section_name())
+                .and_then(|section| section.data().ok())
+                .unwrap_or(&[]);
```
philipc, line 107: I'm going to have to rethink the `object` API here. For compressed sections, we should support either decompressing or returning an error  (based on whether the compression feature is enabled).

src/lib.rs
```diff
@@ -234,7 +237,7 @@ impl<R: gimli::Reader> Context<R> {
             }
 
             res_units.push(ResUnit {
-                dw_unit,
+                dw_unit: Rc::new(dw_unit),
```
philipc, line 240: Is is possible to avoid the `Rc` here?

src/lib.rs
```diff
@@ -285,9 +293,14 @@ impl<R: gimli::Reader> Context<R> {
         Some(unit_id)
     }
 
+    /// Find the DWARF unit corresponding to the given virtual memory address.
+    pub fn find_dw_unit(&self, probe: u64) -> Option<Rc<gimli::Unit<R>>> {
```
philipc, line 297: Please don't use shortened names for public APIs. (It's good to avoid them for internal APIs too, but there's already a lot of that.)
```suggestion
    pub fn find_dwarf_unit(&self, probe: u64) -> Option<Rc<gimli::Unit<R>>> {
```

bjorn3: Done

src/lib.rs
bjorn3: Returned a reference from `find_dwarf_unit` instead.

bjorn3:
CI failure seems unrelated.

philipc:
The coverage failure is unrelated, but the bench failure is due to the object update. When you fix that, can you squash it into the relevant commit?

bjorn3:
Didn't notice the bench failure. Updated it.

PR #160, opened 2020-05-07
Trim dependencies of `addr2line` for `backtrace`

alexcrichton:
This commit trims the dependencies of `addr2line` to be used in the
`backtrace` crate. The purpose here is to slowly trim down the list of
dependencies to make it easier to eventually include this in the
standard library. Fow now this handles three previously required
dependenies of `addr2line`:

* `smallvec` - this is made optional where if not present a standard
  `alloc::vec::Vec<T>` is used.
* `fallible-iterator` - this is simply made optional.
* `lazycell` - this is dropped in favor of a small hand-rolled
  implementation.

The new features are enabled by default and the thinking is that the
`backtrace` crate would disable all these features by default. If the
`std` feature of `backtrace` is activated it'd activate these features,
however.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,10 +16,9 @@ travis-ci = { repository = "gimli-rs/addr2line" }
 
 [dependencies]
 gimli = { version = "0.20", default-features = false, features = ["read"] }
-fallible-iterator = { version = "0.2", default-features = false }
+fallible-iterator = { version = "0.2", default-features = false, optional = true }
 object = { version = "0.18", default-features = false, features = ["read"], optional = true }
-smallvec = { version = "1", default-features = false }
-lazycell = "1"
+smallvec = { version = "1", default-features = false, optional = true }
 rustc-demangle = { version = "0.1", optional = true }
 cpp_demangle = { version = "0.2", default-features = false, optional = true }
 
@@ -37,7 +36,7 @@ debug = true
 codegen-units = 1
 
 [features]
-default = ["rustc-demangle", "cpp_demangle", "std-object"]
+default = ["rustc-demangle", "cpp_demangle", "std-object", "fallible-iterator", "smallvec"]
 std = ["gimli/std"]
 std-object = ["std", "object", "object/std"]
 
--- a/src/lazy.rs
+++ b/src/lazy.rs
@@ -1,0 +1,29 @@
+use core::cell::UnsafeCell;
+
+pub struct LazyCell<T> {
+    contents: UnsafeCell<Option<T>>,
+}
+impl<T> LazyCell<T> {
+    pub fn new() -> LazyCell<T> {
+        LazyCell {
+            contents: UnsafeCell::new(None),
+        }
+    }
+
+    pub fn borrow_with(&self, closure: impl FnOnce() -> T) -> &T {
+        unsafe {
+            // First check if we're already initialized...
+            let ptr = self.contents.get();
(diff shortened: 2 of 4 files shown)


src/lazy.rs
```diff
@@ -0,0 +1,29 @@
...
+            if let Some(val) = &*ptr {
+                return val;
+            }
+            // Note that while we're executing `closure` our `borrow_with` may
+            // be called recursively. This means we need to check again after
+            // the closure has executed. For that we use the `get_or_insert`
+            // method which will only perform mutation if we aren't already
+            // `Some`.
+            let val = closure();
+            (*ptr).get_or_insert(val)
```
bjorn3, line 26:
```suggestion
            (*self.contents.get()).get_or_insert(val)
```

If `borrow_with` is called from inside the closure, then `ptr` is invalidated according to stacked borrows I think.

src/lazy.rs
```diff
@@ -0,0 +1,29 @@
...
+        LazyCell {
+            contents: UnsafeCell::new(None),
+        }
+    }
+
+    pub fn borrow_with(&self, closure: impl FnOnce() -> T) -> &T {
+        unsafe {
+            // First check if we're already initialized...
+            let ptr = self.contents.get();
+            if let Some(val) = &*ptr {
```
bjorn3, line 17:
```suggestion
            if let Some(val) = &*self.contents.get() {
```

Cargo.toml
```diff
@@ -16,10 +16,9 @@ travis-ci = { repository = "gimli-rs/addr2line" }
 
 [dependencies]
 gimli = { version = "0.20", default-features = false, features = ["read"] }
-fallible-iterator = { version = "0.2", default-features = false }
+fallible-iterator = { version = "0.2", default-features = false, optional = true }
```
bjorn3, line 19: This is a required dependency of `gimli` with the `read` feature anyway.

Cargo.toml
```diff
@@ -16,10 +16,9 @@ travis-ci = { repository = "gimli-rs/addr2line" }
...
 object = { version = "0.18", default-features = false, features = ["read"], optional = true }
-smallvec = { version = "1", default-features = false }
-lazycell = "1"
+smallvec = { version = "1", default-features = false, optional = true }
```
bjorn3, line 21: Same for `smallvec`:

https://github.com/gimli-rs/gimli/blob/f346d6fe5434ae3fde26f4c1bf3ff31760690667/Cargo.toml#L39

Cargo.toml
alexcrichton: I, too, am using `cargo tree` and am not simply blindly trying to do things. There's a lot of work that needs to be done [and I'm trying to do it incrementally](https://github.com/gimli-rs/gimli/pull/495).

src/lazy.rs
alexcrichton: I don't believe this is necessary since the value of `ptr` doesn't change, and borrows into it are temporary. If miri or something throws an error here it can of course be changed.

bjorn3: `ptr` is of course a raw pointer.

philipc: @bjorn3 I'm taking your comment to mean that this conversation is resolved.

alexcrichton:
I think a small perf hit is probably ok, but I'll measure one more time before merging all updates to confirn

PR #163, opened 2020-05-12
Don't truncate ranges of units

alexcrichton:
This commit is an attempt to fix rust-lang/backtrace-rs#327. The
behavior seen there was that CGUs would often overlap each other which
would cause a previous loop which tried to ensure CGUs didn't overlap to
avoid some CGUs being candidates for symbolizing.

To fix the issue this commit removes the loop which alters the range of
each CGU. Instead CGUs are only sorted by their starting function
address. When looking up a CGU for a function the CGU's address range is
used to initially filter the sort but afterwards the units function
address table is also consulted. The first CGU which contains the
function address is what ends up getting returned.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -254,23 +254,8 @@ impl<R: gimli::Reader> Context<R> {
             });
         }
 
+        // Sort this for faster lookup in `find_unit_id_and_address` below.
         unit_ranges.sort_by_key(|x| x.0.begin);
-
-        // Ranges need to be disjoint so that we can binary search, but weak symbols can
-        // cause overlap. In this case, we don't care which unit is used, so ignore the
-        // beginning of the subseqent range to avoid overlap.
-        let mut prev_end = 0;
-        for range in &mut unit_ranges {
-            if range.0.begin < prev_end {
-                range.0.begin = prev_end;
-            }
-            if range.0.end < prev_end {
-                range.0.end = prev_end;
-            } else {
-                prev_end = range.0.end;
-            }
-        }
-        debug_assert!(unit_ranges.windows(2).all(|w| w[0].0.end <= w[1].0.begin));
 
         Ok(Context {
             units: res_units,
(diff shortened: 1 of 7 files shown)


alexcrichton:
Ok and now with tests for macos!

src/lib.rs
```diff
@@ -284,35 +269,71 @@ impl<R: gimli::Reader> Context<R> {
...
+        let pos = match self.unit_ranges.binary_search_by_key(&probe, |r| r.0.begin) {
+            // Although unlikely, we could find an exact match.
+            Ok(i) => i,
+            // `probe` is smaller than all CGU start addresses, so it's not
+            // contained anywhere.
+            Err(0) => return None,
+            // No exact match was found, but this probe would fit at slot `i`.
+            // This means that slot `i-1` is the highest start address which is
+            // underneath `probe`, which is the CGU we'd like to start
+            // searching.
```
philipc, line 288: Nit: DWARF calls them compilation units (CU).

Given that CUs can be much larger than needed, it's not clear to me that this is the CU we need to start at. Why can't there be a CU with a lower start address that contains the probe address?

alexcrichton: Hm yes I'm entirely mixed up in how sorting works, let me see if I can find a solution which actually works.

alexcrichton: Ok I think the new logic should work out!

tests/correctness.rs
```diff
@@ -23,24 +49,40 @@ fn correctness() {
...
+            panic!("incorrect name '{}', expected {:?}", name, expected_prefix);
+        }
+    };
+
+    test(test_function as u64, "correctness::test_function");
+    test(
+        small::test_function as u64,
+        "correctness::small::test_function",
+    );
+    test(aux::foo as u64, "aux::foo");
```
philipc, line 70: What are these two tests checking that wasn't covered before?

alexcrichton: The `small` module was me attempting to reproduce the original issue and the `aux::foo` test will fail on macOS before this PR but succeeds after this PR. I'm not entirely sure what makes rustc generate the object that causes this but w/e it is it seems to work.

alexcrichton:
Nah this isn't urgent, just something I noticed. Should be fine to defer a release!

PR #175, opened 2020-05-21
Avoid linear search for DW_FORM_ref_addr

### LINKED ISSUE #171: Slow find_frames on XUL.dSYM, in unit.parse_functions
I'm trying to get inline call stacks for around 60000 addresses from the Firefox-on-Mac `XUL` library, from a `XUL.dSYM` bundle. It takes 2 minutes. That's 2ms per address. [Instruments profile](https://drive.google.com/file/d/1U4DG3nduoHDqwDxRF10GOqmtV12lIOAw/view?usp=sharing)

Doing the same on a Linux libxul.so takes 0.8ms per address, and for PDBs it takes 0.19ms per address.

It would be great if this could be made faster. Once the Firefox profiler supports displaying inline call stacks, I want to use this code to get the inline frame information for local builds, and it would be great to not regress symbolication times too much compared to the current state without inline callstacks.

Steps to reproduce:

```
git clone https://github.com/mstange/profiler-get-symbols/
cd profiler-get-symbols/
git checkout 8152ad4968d847763df3e79388150d87ab214c7e
cargo run -p benchmarks --release
```

Plan around 10-20min for the first run, because it is slow to compile and downloads large resources. Most importantly, this step creates `big-benchmark-fixtures/macos-ci/XUL.dSYM`.

Then run the following to run just the dSYM test (this is the command that obtains inline callstacks for 59736 addresses from XUL.dSYM):

```
cargo run --release -p query-api -- big-benchmark-fixtures/macos-ci/ /symbolicate/v6a1 @fixtures/requests/macos-ci-xul.json > /dev/null
```

Profiling in Instruments reveals that most of the time is spent in `unit.parse_functions`. It seems that if many addresses fall into the same unit (?) we might be doing a lot of duplicated work.

Screen Shot 2020-05-18 at 8 34 11 PM
Screen Shot 2020-05-18 at 8 38 10 PM

philipc:
Fixes #171 (cc @mstange)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,6 +189,7 @@ impl<R: gimli::Reader> Context<R> {
         let mut units = sections.units();
         while let Some(header) = units.next()? {
             let unit_id = res_units.len();
+            let offset = header.offset();
             let dw_unit = match sections.unit(header) {
                 Ok(dw_unit) => dw_unit,
                 Err(_) => continue,
@@ -257,6 +258,7 @@ impl<R: gimli::Reader> Context<R> {
             }
 
             res_units.push(ResUnit {
+                offset,
                 dw_unit,
                 lang,
                 lines: LazyCell::new(),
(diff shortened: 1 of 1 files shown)


jdm:
Could a new release be published with this change? I suspect this will help the performance of generating backtraces for Servo as well.

philipc:
@jdm Published 0.12.2

PR #176, opened 2020-05-23
Improve performance by parsing function contents lazily

mstange:
This is a somewhat major rework of when and how functions are parsed and what representation they're stored in.

On my benchmarks, it improves overall `find_frames` performance by 1.8x to 3x, or, in absolute numbers, from 25-40 seconds to 10-20 seconds. The benchmarks in this crate are improved as well, with two caveats (see further down).

These changes are motivated by these insights:
 1. On current master, there is already one level of laziness: The functions in a unit are only parsed once we encounter an address in that unit. However, in practice, the addresses that are looked up only fall into a small subset of those functions, so we're parsing the contents of many functions unnecessarily. Moreover, this existing level of laziness is not effective in the split debug info scenario, where there's only one unit per object file to begin with.
 2. Parsing names of functions is slow. It takes up memory bandwidth, and, when the data is coming from an mmap'd file, causes more parts of the file to mapped into memory. Parsing function names should be avoided unless the name is actually needed.
 3. The vast majority of DW_TAG_subprogram instances don't cover an address range. They're functions from header files, which are repeated as DW_TAG_inlined_subroutine for the address ranges they actually end up covering in the binary. We were parsing their names unnecessarily.

This PR does the following:
 - There are now two levels of function parsing laziness.
   - The first level creates only the function ranges for "outer functions" (DW_TAG_subprogram) and writes down each outer function's dw_die_offset. By necessity, when parsing the list, we need to iterate over all abbrevs and attributes of the unit, including the ones for any nested tags for inlined functions, but we try to keep the work we do for them to a minimum.
   - The second level is done once we look up an address that falls into an outer function's address range. At that point, we parse the entries for that outer function again, starting at the recorded dw_die_offset. We parse the function name and all functions that are inlined into it, with their names and other data.
 - This two-level laziness automatically makes it so that we don't parse the names of DW_TAG_subprogram instances that don't have address ranges.
 - Reduced allocations for inline functions: For nested inline functions, we no longer have a Vec of child inlines on each inline function. Instead, we now have one Vec with all descendant inline_address_ranges on the OuterFunction.
 - We do a binary search when looking up the inline functions for an address within an outer function.

`cargo bench` results:

```
% cargo benchcmp bench-master.txt bench-better.txt
 name                                        bench-master.txt ns/iter  bench-better.txt ns/iter  diff ns/iter   diff %  speedup 
 context_new_and_query_location_rc           4,726,113                 3,377,543                   -1,348,570  -28.53%   x 1.40 
 context_new_and_query_location_slice        2,003,584                 1,542,540                     -461,044  -23.01%   x 1.30 
 context_new_and_query_with_functions_rc     4,668,415                 4,031,275                     -637,140  -13.65%   x 1.16 
 context_new_and_query_with_functions_slice  1,953,032                 1,570,379                     -382,653  -19.59%   x 1.24 
 context_new_parse_functions_rc              69,957,646                26,982,504                 -42,975,142  -61.43%   x 2.59 
 context_new_parse_functions_slice           50,998,880                20,554,205                 -30,444,675  -59.70%   x 2.48 
 context_new_parse_lines_rc                  17,311,785                17,292,187                     -19,598   -0.11%   x 1.00 
 context_new_parse_lines_slice               11,412,235                11,999,761                     587,526    5.15%   x 0.95 
 context_new_rc                              3,192,877                 3,245,348                       52,471    1.64%   x 0.98 
 context_new_slice                           1,052,836                 1,054,370                        1,534    0.15%   x 1.00 
 context_query_location_rc                   1,126,921                 1,130,779                        3,858    0.34%   x 1.00 
 context_query_location_slice                1,171,425                 1,142,100                      -29,325   -2.50%   x 1.03 
 context_query_with_functions_rc             3,856,426                 3,803,931                      -52,495   -1.36%   x 1.01 
 context_query_with_functions_slice          3,610,648                 3,452,011                     -158,637   -4.39%   x 1.05 
```

The two important improvements are the ones in `context_new_and_query_with_functions`:

```
 context_new_and_query_with_functions_rc      -13.65%   x 1.16 
 context_new_and_query_with_functions_slice   -19.59%   x 1.24 
```

Caveats:
 1. The `parse_functions` tests now have a different meaning, so their improvement is artificial. `parse_functions` longer parses all functions, it now only parses the list of outer functions.
 2. There are some tests that show worse numbers now, especially `context_new_parse_lines_slice`. This is confusing to me because I don't think I actually changed what happens in them. I think these numbers are a bit flaky and are affected by the file system cache to a large degree. I don't think I actually made anything slower.

I apologize for the size of this PR. The commits aren't the cleanest; some of them don't compile and I sometimes go back and forth between different solutions. The first commit is reverted in the last commit.
If you'd like, I can rewrite the PR as a set with more meaningfully grouped commits. Please let me know if that's necessary. Also, please feel free to make any edits you like.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,26 +205,11 @@ impl<R: gimli::Reader> Context<R> {
                     _ => continue, // wtf?
                 };
 
-                let mut low_pc = None;
-                let mut high_pc = None;
-                let mut size = None;
-                let mut ranges = None;
+                let mut range_components: RangeComponents<R> = Default::default();
                 for spec in abbrev.attributes() {
                     let attr = entries.read_attribute(*spec)?;
+                    range_components.parse_range_attr(&attr, &sections, &dw_unit)?;
                     match attr.name() {
-                        gimli::DW_AT_low_pc => {
-                            if let gimli::AttributeValue::Addr(val) = attr.value() {
-                                low_pc = Some(val);
-                            }
-                        }
-                        gimli::DW_AT_high_pc => match attr.value() {
-                            gimli::AttributeValue::Addr(val) => high_pc = Some(val),
-                            gimli::AttributeValue::Udata(val) => size = Some(val),
-                            _ => {}
-                        },
-                        gimli::DW_AT_ranges => {
-                            ranges = sections.attr_ranges_offset(&dw_unit, attr.value())?;
-                        }
                         gimli::DW_AT_language => {
                             if let gimli::AttributeValue::Language(val) = attr.value() {
                                 lang = Some(val);
@@ -234,28 +219,13 @@ impl<R: gimli::Reader> Context<R> {
                     }
                 }
 
-                let mut add_range = |range: gimli::Range| {
-                    if range.begin < range.end {
-                        unit_ranges.push(UnitRange {
-                            range,
-                            unit_id,
-                            max_end: 0,
-                        });
-                    }
-                };
-                if let Some(offset) = ranges {
-                    let mut ranges = sections.ranges(&dw_unit, offset)?;
-                    while let Some(range) = ranges.next()? {
-                        add_range(range);
-                    }
-                } else if let (Some(begin), Some(end)) = (low_pc, high_pc) {
-                    add_range(gimli::Range { begin, end });
-                } else if let (Some(begin), Some(size)) = (low_pc, size) {
-                    add_range(gimli::Range {
-                        begin,
-                        end: begin + size,
+                range_components.for_each_range(&sections, &dw_unit, |range| {
+                    unit_ranges.push(UnitRange {
+                        range,
+                        unit_id,
+                        max_end: 0,
                     });
-                }
+                })?;
             }
 
             res_units.push(ResUnit {
@@ -782,24 +752,19 @@ impl<R: gimli::Reader> Functions<R> {
                     for spec in abbrev.attributes() {
                         match entries.read_attribute(*spec) {
                             Ok(ref attr) => {
-                                Function::parse_range_attr(
-                                    attr,
-                                    &mut range_components,
-                                    sections,
-                                    unit,
-                                )?;
+                                range_components.parse_range_attr(attr, sections, unit)?;
                             }
                             Err(e) => return Err(e),
                         }
                     }
 
                     let function_index = functions.len();
-                    let have_ranges = range_components.add_ranges(
-                        &mut function_address_ranges,
-                        function_index,
-                        sections,
-                        unit,
-                    )?;
+                    let have_ranges = range_components.for_each_range(sections, unit, |range| {
+                        function_address_ranges.push(FunctionAddressRange {
+                            range,
+                            function: function_index,
+                        });
+                    })?;
                     if have_ranges {
                         functions.push(Function {
                             dw_die_offset,
@@ -908,12 +873,36 @@ impl<R: gimli::Reader> Default for RangeComponents<R> {
 }
 
 impl<R: gimli::Reader> RangeComponents<R> {
-    fn add_ranges(
-        &self,
-        ranges: &mut Vec<FunctionAddressRange>,
-        function: usize,
+    fn parse_range_attr(
+        &mut self,
+        attr: &gimli::read::Attribute<R>,
         sections: &gimli::Dwarf<R>,
         unit: &gimli::Unit<R>,
+    ) -> Result<(), Error> {
+        match attr.name() {
+            gimli::DW_AT_low_pc => {
+                if let gimli::AttributeValue::Addr(val) = attr.value() {
+                    self.low_pc = Some(val);
+                }
+            }
+            gimli::DW_AT_high_pc => match attr.value() {
+                gimli::AttributeValue::Addr(val) => self.high_pc = Some(val),
+                gimli::AttributeValue::Udata(val) => self.size = Some(val),
+                _ => {}
+            },
+            gimli::DW_AT_ranges => {
+                self.range_list_offset = sections.attr_ranges_offset(unit, attr.value())?;
+            }
+            _ => {}
+        };
+        Ok(())
+    }
+
+    fn for_each_range<F: FnMut(gimli::Range)>(
+        &self,
+        sections: &gimli::Dwarf<R>,
+        unit: &gimli::Unit<R>,
+        mut f: F,
     ) -> Result<bool, Error> {
         let mut added_any = false;
         let mut add_range = |range: gimli::Range| {
@@ -921,7 +910,7 @@ impl<R: gimli::Reader> RangeComponents<R> {
             // a long list of matches.
             // TODO: don't ignore if there is a section at this address
             if range.begin != 0 && range.begin < range.end {
-                ranges.push(FunctionAddressRange { range, function });
+                f(range);
                 added_any = true
             }
         };
@@ -1022,7 +1011,7 @@ impl<R: gimli::Reader> Function<R> {
         for spec in abbrev.attributes() {
             match entries.read_attribute(*spec) {
                 Ok(ref attr) => {
-                    Self::parse_range_attr(attr, &mut range_components, sections, unit)?;
+                    range_components.parse_range_attr(attr, sections, unit)?;
                     Self::parse_name_attr(attr, &mut name, sections, unit, units)?;
 
                     match attr.name() {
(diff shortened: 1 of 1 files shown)


mstange:
After these changes, profiling my `profiler-get-symbols` benchmarks gives me the following breakdowns:

Breakdown of XUL.dSYM benchmark:
 - 35% parsing outer functions list of a unit
 - 30% gimli parsing a unit into list of abbreviations + attributes
 - 10% deallocating context + dwarf data
 - 10% copying data out of addr2lines frame iterator
 - 7.5% parsing an outer function with its inlines
 - 5% parsing lines (excluding render_file)
 - 1.5% render_file

Breakdown for local macOS build with split debug info:
 - 80% parsing outer functions list of a unit
 - 9% parsing an outer function with its inlines
 - 4.5% parsing lines (excluding render_file)
 - 1.5% render_file
 - 1.4% gimli parsing a unit's list of abbrevs + attributes
 - 0.3% dealloc

In the XUL.dSYM case, the biggest potential for improvement would be to have gimli stop allocating its own representation of abbrev + attribute lists, or build them more lazily.

In the split debug info case, there's not much room for improvement, unless there is a way to get the dw_die_offsets of all the subprogram tags without having to iterate over (and parse) all the abbrevs + attributes of any nested DIEs.

philipc:
These changes sound great! Thanks for doing this.

> If you'd like, I can rewrite the PR as a set with more meaningfully grouped commits. Please let me know if that's necessary.

It'd be great if you can do that. It will make it easier for me to review.

> The parse_functions tests now have a different meaning, so their improvement is artificial. parse_functions longer parses all functions, it now only parses the list of outer functions.

It'd be good if we had a benchmark for the inner function parsing too, but I can add that later if you want. We won't be able to compare it against the performance in master.

> There are some tests that show worse numbers now, especially context_new_parse_lines_slice. This is confusing to me because I don't think I actually changed what happens in them. I think these numbers are a bit flaky and are affected by the file system cache to a large degree. I don't think I actually made anything slower.

Yeah there's a lot of noise in the tests. Running them multiple times might help get stabler numbers.

> Breakdown of XUL.dSYM benchmark

This is the one you gave in #171?

> Breakdown for benchmark that looks up addresses of a local Firefox macOS build with split debug info:

Do you have any insight into why these are so different? My intuition was that they would be the same, since the dSYM is just all the objects bundled together? Is there any way I can do a similar performance measurement without needing a local Firefox macOS build?

> In the XUL.dSYM case, the biggest potential for improvement would be to have gimli stop allocating its own representation of abbrev + attribute lists, or build them more lazily.

I doubt that building them lazily will help unless using `DW_AT_sibling` lets you skip a significant number of them.

bjorn3:
> My intuition was that they would be the same, since the dSYM is just all the objects bundled together?

A `dSYM` only contains a single object file that contains the merged debuginfo of all linked object files together. Also a `dSYM` doesn't use any debuginfo in the linked executable/library to match functions. Instead it contains the exact offset of every function. DWARF split debuginfo however uses debuginfo in the linked executable/library to match function locations and debuginfo in the split file.

philipc:
> A dSYM only contains a single object file that contains the merged debuginfo of all linked object files together. Also a dSYM doesn't use any debuginfo in the linked executable/library to match functions. Instead it contains the exact offset of every function.

I know the dSYM is a single Mach-O file containing only the debuginfo, but why does that affect performance? We're still only parsing the debuginfo sections in both cases. Isn't the merged debuginfo effectively just the concatenated and relocated `.debug_info`/`.debug_line`/etc sections from the object files? If so, why does it make a difference if the debuginfo is in many object files versus contained in a single dSYM?

> DWARF split debuginfo however uses debuginfo in the linked executable/library to match function locations and debuginfo in the split file.

What's the relevancy of that? I don't think either of those performance breakdowns were for DWARF split debuginfo?

bjorn3:
> What's the relevancy of that? I don't think either of those performance breakdowns were for DWARF split debuginfo?

The first breakdown was for dSYM, the second for DWARF split debuginfo I think.

philipc:
Ah I see. I'm pretty sure macOS doesn't use DWARF split debuginfo, but I could be wrong. I think split debuginfo in the context of that breakdown means using the object files prior to running dsymutil (they are located via the symbol table).

mstange:
Sorry, yes, when I said "split debuginfo" I meant "debuginfo in object files referenced via OSO stab symbols in the XUL library". What's the right name for that?

> If so, why does it make a difference if the debuginfo is in many object files versus contained in a single dSYM?

My guess is that we're just not parsing many of those object files, because they don't contain any of the addresses that we're looking up. So in the dSYM case, when we parse the list of units, we need to iterate over all units in the entire dSYM. But in the object file case, we're only encountering the units in the object files that we know are relevant. So in effect, the OSO symbols act as an index for the relevant dwarf units.

mstange:
> These changes sound great! Thanks for doing this.
> 
> > If you'd like, I can rewrite the PR as a set with more meaningfully grouped commits. Please let me know if that's necessary.
> 
> It'd be great if you can do that. It will make it easier for me to review.

Sure, I'll try to find some time for that. Shouldn't be too much work now that I know what I want the outcome to be.

> > The parse_functions tests now have a different meaning, so their improvement is artificial. parse_functions longer parses all functions, it now only parses the list of outer functions.
> 
> It'd be good if we had a benchmark for the inner function parsing too, but I can add that later if you want. We won't be able to compare it against the performance in master.

Agreed.

> > Breakdown of XUL.dSYM benchmark
> 
> This is the one you gave in #171?

That's right.

> > Breakdown for benchmark that looks up addresses of a local Firefox macOS build with split debug info:
> 
> Do you have any insight into why these are so different? My intuition was that they would be the same, since the dSYM is just all the objects bundled together?

(answered in previous comment)

> Is there any way I can do a similar performance measurement without needing a local Firefox macOS build?

I'm working on that. I've uploaded an archive with all the necessary object files to Google Cloud storage now, so I should be able to add a benchmark that uses it.

> > In the XUL.dSYM case, the biggest potential for improvement would be to have gimli stop allocating its own representation of abbrev + attribute lists, or build them more lazily.
> 
> I doubt that building them lazily will help unless using `DW_AT_sibling` lets you skip a significant number of them.

A big part of the profile is in the allocator. I haven't looked into it too deeply but it looks like gimli is building Vecs with lots of information during this first parse. Sure, we may have to iterate over everything once at the beginning, but maybe it's possible to just create an index with some offsets during the first parse. And then a second parse can be done lazily once the relevant unit is needed. Analogous to what I'm doing with functions in this PR.

mstange:
> [...]  but maybe it's possible to just create an index with some offsets during the first parse. 

Oh, we also need the ranges for each unit. So I suppose the real challenge is: When building the initial list of units, get the pc range or range list of a unit without allocating any vectors for that unit. Or something like that.

philipc:
> Oh, we also need the ranges for each unit. So I suppose the real challenge is: When building the initial list of units, get the pc range or range list of a unit without allocating any vectors for that unit. Or something like that.

Ah yep, that's something we could optimize. At that stage we only need the abbreviation for the compilation unit DIE.  This is https://github.com/gimli-rs/gimli/issues/251, but the trait solution I describe there wouldn't be needed now that we're using lower level DIE parsing here.

philipc:
Another thing that we should do is `.debug_aranges` support, which would achieve a similar goal.

philipc:
As a suggestion, I could do the rebase myself and get you to review any changes I make?

mstange:
I think I'll have time to work on it this coming weekend.

philipc:
It seems you have little time to work on this, and I do have spare time currently. I've started doing the rebase, and I think it will be a worthwhile exercise, but let me know if that will be a problem.

One thing I would like to do is to see if the `RefCell` and `Rc` can be avoided (using `LazyCell` instead like the rest of the code). Conceptually I think this should be possible.

Also I see that you temporarily had a `LocationOnly` state for `FrameIter`. This actually seems like something we should need, and it may be a bug that we're not handling this correctly currently. That is, it should be possible to handle debuginfo that has line number information but not function inlining information.

mstange:
> It seems you have little time to work on this, and I do have spare time currently. I've started doing the rebase, and I think it will be a worthwhile exercise, but let me know if that will be a problem.

Sorry, yes, that sounds great.

> One thing I would like to do is to see if the `RefCell` and `Rc` can be avoided (using `LazyCell` instead like the rest of the code). Conceptually I think this should be possible.

Definitely worth trying! This was actually my first time writing code that uses RefCell and I wasn't entirely sure if it was a good choice. But conceptually it does make sense to me to have shared ownership of the function data between the iterator and the cache.

> Also I see that you temporarily had a `LocationOnly` state for `FrameIter`. This actually seems like something we should need, and it may be a bug that we're not handling this correctly currently. That is, it should be possible to handle debuginfo that has line number information but not function inlining information.

Sounds good. I think I added it when I thought "oh no I broke this case", only to later realize that the original code didn't handle it either (I think), so I removed it again because I didn't want to introduce behavioral differences in an optimization patch.

philipc:
> Sounds good. I think I added it when I thought "oh no I broke this case", only to later realize that the original code didn't handle it either (I think), so I removed it again because I didn't want to introduce behavioral differences in an optimization patch.

I didn't pursue this further. I agree that would be better for a new patch.

PR #178, opened 2020-06-28
Improve performance by parsing function contents lazily

philipc:
This is a significant rebase of the great work done by @mstange in #176 . Closes #176 .

The description from that PR still applies:

>    - There are now two levels of function parsing laziness.
>        - The first level creates only the function ranges for "outer functions" (DW_TAG_subprogram) and writes down each outer function's dw_die_offset. By necessity, when parsing the list, we need to iterate over all abbrevs and attributes of the unit, including the ones for any nested tags for inlined functions, but we try to keep the work we do for them to a minimum.
>        - The second level is done once we look up an address that falls into an outer function's address range. At that point, we parse the entries for that outer function again, starting at the recorded dw_die_offset. We parse the function name and all functions that are inlined into it, with their names and other data.
>    - This two-level laziness automatically makes it so that we don't parse the names of DW_TAG_subprogram instances that don't have address ranges.
>    - Reduced allocations for inline functions: For nested inline functions, we no longer have a Vec of child inlines on each inline function. Instead, we now have one Vec with all descendant inline_address_ranges on the OuterFunction.
>    - We do a binary search when looking up the inline functions for an address within an outer function.

Differences from that PR:
- cleaner commit history
- new benchmarks
- use `LazyCell` instead of `Rc`/`RefCell`, since this makes the code easier to understand
- less factoring out of attribute parsing, since this slightly affects performance
- cosmetic changes to minimize differences to master

This PR was written by completely rewriting the commits instead of doing a normal rebase. Hopefully I haven't left out anything significant.

@mstange Could you please review this, and check against the benchmarks you were using? And thanks again for your work on this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -324,7 +324,7 @@ impl<R: gimli::Reader> Context<R> {
             // if there's actually a function in this CU which contains this
             // address do we return this unit.
             let unit = &self.units[i.unit_id];
-            let funcs = match unit.parse_functions(&self.sections, &self.units) {
+            let funcs = match unit.parse_functions(&self.sections) {
                 Ok(func) => func,
                 Err(_) => continue,
             };
@@ -366,9 +366,14 @@ impl<R: gimli::Reader> Context<R> {
         };
 
         let loc = unit.find_location(probe, &self.sections)?;
-        let functions = unit.parse_functions(&self.sections, &self.units)?;
+        let functions = unit.parse_functions(&self.sections)?;
         let function_index = functions.addresses[address].function;
         let function = &functions.functions[function_index];
+        let function = function
+            .1
+            .borrow_with(|| Function::parse(function.0, &unit.dw_unit, &self.sections, &self.units))
+            .as_ref()
+            .map_err(Error::clone)?;
 
         // Build the list of inlined functions that contain `probe`.
         // `inlined_functions` is ordered from outside to inside.
@@ -424,7 +429,7 @@ impl<R: gimli::Reader> Context<R> {
     #[doc(hidden)]
     pub fn parse_functions(&self) -> Result<(), Error> {
         for unit in &self.units {
-            unit.parse_functions(&self.sections, &self.units)?;
+            unit.parse_functions(&self.sections)?;
         }
         Ok(())
     }
@@ -433,7 +438,7 @@ impl<R: gimli::Reader> Context<R> {
     #[doc(hidden)]
     pub fn parse_inlined_functions(&self) -> Result<(), Error> {
         for unit in &self.units {
-            unit.parse_functions(&self.sections, &self.units)?;
+            unit.parse_inlined_functions(&self.sections, &self.units)?;
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


bjorn3:
You can use `Co-authored-by:` https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

mstange:
I'm reviewing this now.

src/lib.rs
```diff
@@ -376,38 +360,62 @@ impl<R: gimli::Reader> Context<R> {
...
+        let function_index = functions.addresses[address].function;
+        let function = &functions.functions[function_index];
+        let function = function
+            .1
+            .borrow_with(|| Function::parse(function.0, &unit.dw_unit, &self.sections, &self.units))
+            .as_ref()
+            .map_err(Error::clone)?;
+
+        // Build the list of inlined functions that contain probe.
+        // funcs is ordered from outside to inside.
```
mstange, line 379: `funcs` should be `inlined_functions`

src/lib.rs
```diff
@@ -376,38 +360,62 @@ impl<R: gimli::Reader> Context<R> {
...
+        // The indexes are into the function.inline_functions vec.
```
mstange, line 380: I'm not sure which "indexes" this comment refers to. Looking at my PR, the comment looked just as lost in it. I think it can be removed.

src/lib.rs
```diff
@@ -376,38 +360,62 @@ impl<R: gimli::Reader> Context<R> {
...
+        let mut inlined_functions = maybe_small::Vec::new();
+        // Walk down the path of inline functions that contain probe.
+        let mut inlined_addresses = &function.inlined_addresses[..];
+        loop {
+            let current_depth = inlined_functions.len();
+            // Look up (probe, current_depth) in inline_ranges.
+            // inline_ranges is sorted in "breadth-first traversal order", i.e.
```
mstange, line 387: `inline_ranges` should be `inlined_addresses`

src/lib.rs
```diff
@@ -533,15 +550,23 @@ where
...
         sections: &gimli::Dwarf<R>,
         units: &[ResUnit<R>],
-    ) -> Result<&Functions<R>, Error> {
+    ) -> Result<(), Error> {
         self.funcs
-            .borrow_with(|| Functions::parse(&self.dw_unit, sections, units))
+            .borrow_with(|| Functions::parse(&self.dw_unit, sections))
             .as_ref()
-            .map_err(Error::clone)
+            .map_err(Error::clone)?
```
mstange, line 568: Oh good, this is better than what I did because you store the error, rather than re-parsing errored functions on every call.

src/lib.rs
```diff
@@ -860,114 +917,300 @@ impl<R: gimli::Reader> Function<R> {
...
-                    )?;
-                } else {
-                    for spec in abbrev.attributes() {
-                        match entries.read_attribute(*spec) {
-                            Ok(_) => {}
-                            Err(e) => return Err(e),
+                match abbrev.tag() {
+                    gimli::DW_TAG_subprogram => {
+                        Function::skip(entries, abbrev, next_depth)?;
+                    }
```
mstange, line 990: Interesting. Yeah, I think you're handling this more correctly than I was. I think my PR was siphoning inlines from nested subprograms into the outer subprogram.

src/lib.rs
philipc: I think I changed the code to return a reference instead of the index.

Issue #190, opened 2020-09-16
Filename is empty with older DWARF version


viveksjain:
I hit a curious issue where the line number was returned correctly yet filename was empty. From what I can gather based on debugging, this happens because of the logic in https://github.com/gimli-rs/addr2line/blob/424c80f7b658455accaa85dd17c2d76515c1fa6c/src/lib.rs#L495 combined with https://github.com/gimli-rs/gimli/blob/e046ca50e2d16ebe775a798d20ad947b7430859f/src/read/line.rs#L1246 - for my binary, comp_file ends up being empty and thus the loop terminates early. This may or may not be caused by the fact that my binary is generated with split-dwarf.

The following patch fixes the issue for me but I am not sure if it is the right fix:
```diff
diff --git a/src/lib.rs b/src/lib.rs
index a03630c..e40e485 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -490,8 +490,12 @@ where
                 sequences.sort_by_key(|x| x.start);

                 let mut files = Vec::new();
-                let mut index = 0;
                 let header = ilnp.header();
+                match header.file(0) {
+                    Some(file) => files.push(self.render_file(file, header, sections)?),
+                    None => files.push(String::from("")),  // DWARF version <= 4 may not have 0th index
+                }
+                let mut index = 1;
                 while let Some(file) = header.file(index) {
                     files.push(self.render_file(file, header, sections)?);
                     index += 1;
```

philipc:
Thanks for the report! Are you able to give some minimal steps for how to reproduce this? I can probably work out how, but if you already know then that's easier.

viveksjain:
I'm afraid I can't share the binary. I'm guessing this happens because of split dwarf (and multiple source/object files). I don't have commands handy to generate an example, but if you want I can experiment to try to create one.

philipc:
It's ok, I've reproduced it now.

philipc:
While there's more we need to do to support split dwarf, I think that fix is a good step. Would you like to submit a PR with it?

bjorn3:
According to the docs index 0 is not valid for DWARF <= 4. addr2line should start iterating at 1 for DWARF <= 4.

philipc:
The version check is already done in gimli (and you could argue that gimli should always return `None` for index 0 and DWARF <=4, but that's a separate issue).

viveksjain:
[...]

PR #191, opened 2020-09-16
Handle older DWARF versions better

viveksjain:
DWARF version <= 4 may not have index 0. As a workaround, we add a dummy
value, which should be fine since presumably 0 index should not be
referenced in any valid DWARF file <= v4. Fixes gimli-rs/addr2line#190.


philipc (approved):
Thanks!

PR #200, opened 2020-12-19
Use .debug_aranges

philipc:
Some compilers only emit the full ranges in .debug_aranges,
leaving only low/high attributes on the CU, which may overlap
with other CUs.  We already have workarounds for this, but it's
better to use the ranges if we have them.


philipc:
Needs https://github.com/gimli-rs/gimli/pull/539

PR #204, opened 2021-01-27
Update object.

dvc94ch:


philipc:
Thanks for the PR. `object` now needs at least rust 1.42, so that will need updating here too. Also, I'll probably do another `object` release before releasing a new `addr2line` version, so just curious if there was a reason you needed this updated now?

dvc94ch:
addr2line reexports `object`, so it's not really pressing

PR #207, opened 2021-02-19
Bump `object` to `0.23.0`

### LINKED ISSUE #206: Bump `object` to `0.23.0`
Please bump the dependency `object` to `0.23.0` 😄

Urhengulas:
Fixes #206

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "gimli-rs/addr2line" }
 [dependencies]
 gimli = { version = "0.23", default-features = false, features = ["read"] }
 fallible-iterator = { version = "0.2", default-features = false, optional = true }
-object = { version = "0.22", default-features = false, features = ["read"], optional = true }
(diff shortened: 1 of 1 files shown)


philipc:
Thanks for the PR. `object` now needs at least rust 1.42, so that will need updating here too. Also, I'll probably do another `object` release before releasing a new `addr2line` version, so just curious if there was a reason you needed this updated now?

Urhengulas:
> Thanks for the PR. `object` now needs at least rust 1.42, so that will need updating here too. Also, I'll probably do another `object` release before releasing a new `addr2line` version, so just curious if there was a reason you needed this updated now?

There is actually no pressing need. The main motivation is keep the dependencies of [`defmt`](https://github.com/knurling-rs/probe-rundefmt) and [`probe-run`](https://github.com/knurling-rs/probe-run) up to date, but also avoid having multiple versions of one dependencies in the tree.

philipc:
I'll do the bump and CI fix in #208, but the plan is still to not do a release until after another `object` release.

PR #209, opened 2021-02-25
Added windows paths support

### LINKED ISSUE #43: Fix logic to handle absolute paths on Windows
Our logic for absolute paths looks for a leading '/', but this doesn't work on Windows.

diondokter:
Fixes #43 

I don't know if this is the best way to do it, but it seems to work.
Normally I would use the std for this, but apparently this is a no_std crate. I guess for binary size?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -892,14 +892,32 @@ impl<'ctx> Iterator for LocationRangeUnitIter<'ctx> {
 }
 
 fn path_push(path: &mut String, p: &str) {
-    if p.starts_with('/') {
+    if has_unix_root(p) || has_windows_root(p) {
         *path = p.to_string();
     } else {
-        if !path.ends_with('/') {
-            path.push('/');
+        let dir_separator = match (has_unix_root(path.as_str()), has_windows_root(path.as_str())) {
+            (true, _) => '/', // Unix path
+            (false, true) => '\\', // Windows path
+            (false, false) => '/', // Unknown path, so let's do it the unix way
+        };
+
+        if !path.ends_with(dir_separator) {
+            path.push(dir_separator);
         }
         *path += p;
     }
+}
+
+/// Check if the path in the given string has a unix style root
+fn has_unix_root(p: &str) -> bool {
+    p.starts_with('/')
+}
+
+/// Check if the path in the given string has a windows style root
+fn has_windows_root(p: &str) -> bool {
+    p.starts_with('\\')
+    || p.get(0..2).map(|p| p == "\\\\").unwrap_or(false)
+    || p.get(1..3).map(|p| p == ":\\").unwrap_or(false)
 }
 
 fn name_attr<R>(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -895,11 +895,7 @@ fn path_push(path: &mut String, p: &str) {
     if has_unix_root(p) || has_windows_root(p) {
         *path = p.to_string();
     } else {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -892,16 +892,34 @@ impl<'ctx> Iterator for LocationRangeUnitIter<'ctx> {
...
+    if has_unix_root(p) || has_windows_root(p) {
         *path = p.to_string();
     } else {
-        if !path.ends_with('/') {
-            path.push('/');
+        let dir_separator = match (has_unix_root(path.as_str()), has_windows_root(path.as_str())) {
+            (true, _) => '/', // Unix path
+            (false, true) => '\\', // Windows path
+            (false, false) => '/', // Unknown path, so let's do it the unix way
+        };
```
philipc, line 902: Can simplify this to: `if has_windows_root(path.as_str()) { '\\' } else { '/' }`.

src/lib.rs
```diff
@@ -892,16 +892,34 @@ impl<'ctx> Iterator for LocationRangeUnitIter<'ctx> {
...
 
+/// Check if the path in the given string has a unix style root
+fn has_unix_root(p: &str) -> bool {
+    p.starts_with('/')
+}
+
+/// Check if the path in the given string has a windows style root
+fn has_windows_root(p: &str) -> bool {
+    p.starts_with('\\')
+    || p.get(0..2).map(|p| p == "\\\\").unwrap_or(false)
```
philipc, line 919: This seems redundant given the first check?

src/lib.rs
```diff
@@ -892,16 +892,34 @@ impl<'ctx> Iterator for LocationRangeUnitIter<'ctx> {
...
     } else {
-        if !path.ends_with('/') {
-            path.push('/');
+        let dir_separator = match (has_unix_root(path.as_str()), has_windows_root(path.as_str())) {
+            (true, _) => '/', // Unix path
+            (false, true) => '\\', // Windows path
+            (false, false) => '/', // Unknown path, so let's do it the unix way
+        };
+
+        if !path.ends_with(dir_separator) {
```
philipc, line 904: Maybe we should allow either separator here? I've seen mingw tools generate paths that include both separators.

src/lib.rs
```diff
@@ -892,16 +892,34 @@ impl<'ctx> Iterator for LocationRangeUnitIter<'ctx> {
...
+/// Check if the path in the given string has a unix style root
+fn has_unix_root(p: &str) -> bool {
+    p.starts_with('/')
+}
+
+/// Check if the path in the given string has a windows style root
+fn has_windows_root(p: &str) -> bool {
+    p.starts_with('\\')
+    || p.get(0..2).map(|p| p == "\\\\").unwrap_or(false)
+    || p.get(1..3).map(|p| p == ":\\").unwrap_or(false)
```
philipc, line 920:
```suggestion
    || p.get(1..3) == Some(":\\")
```

philipc (changes requested):
> I don't know if this is the best way to do it, but it seems to work.

Seems good to me!

> Normally I would use the std for this, but apparently this is a no_std crate. I guess for binary size?

This crate is used in `backtrace-rs`, which is used to implement backtraces in `libstd`, so it can't depend on `libstd` for that.

src/lib.rs
philipc: Hmm, that seems to just be rust's prefix map which looks wrong (e.g. `/rustc/4fb7144ed159f94491249e86d5bbd033b5d60550\src\libcore\str`). This is probably fine how it is.

philipc (approved):

Issue #106, opened 2018-11-27
Trait bound not satisfied


olson-dan:
Apologies if this has an obvious fix.

Making a new crate as so:
> cargo new test_crate
> cd test_crate
> cargo add object
> cargo add addr2line

Using this source:
```rust
extern crate object;
extern crate addr2line;

use addr2line::Context;
use object::Object;

fn main() {
    let data: Vec<u8> = Vec::new();
    let obj = object::File::parse(&data).unwrap();
    let addr2line = Context::new(&obj).unwrap();
    println!("Hello World");
}
```

I get the following compile error, even though I believe this should work based on the README and examples:

```
warning: unused import: `object::Object`
 --> src\main.rs:5:5
  |
5 | use object::Object;
  |     ^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

error[E0277]: the trait bound `object::File<'_>: object::traits::Object<'_, '_>` is not satisfied
  --> src\main.rs:10:21
   |
10 |     let addr2line = Context::new(&obj).unwrap();
   |                     ^^^^^^^^^^^^ the trait `object::traits::Object<'_, '_>` is not implemented for `object::File<'_>`
   |
   = note: required by `<addr2line::Context<gimli::endian_reader::EndianReader<gimli::endianity::RunTimeEndian, std::rc::Rc<[u8]>>>>::new`
```

The versions being used here are `object = "0.11.0"` and `addr2line = "0.7.0"`.  I'm not sure why the trait bound is not satisfied, because everything looks fine according to the documentation and `object::File` correctly implements `object::Object`.  Also I pulled in `object::Object` with `use` because I thought that might have been the issue, but all this did was give me a warning.  I'm using rustc 1.30.1 on stable-x86_64-pc-windows-msvc

olson-dan:
The example here has the same issue as my test case: https://github.com/gimli-rs/addr2line/blob/master/examples/addr2line.rs

philipc:
Yes, you need to use the version of object that is depended on by addr2line (0.10.0). This is a general problem in rust when types from other crates are exposed as part of the public API, and the error messages aren't very helpful when this happens.

tmplt:
Note that the correct version of `object` is exported by `addr2line` under `addr2line::object`. So
```rust
use addr2line;

// ...

let obj = addr2line::object::File::parse(&data).unwrap();
let ctx = addr2line::Context::new(&obj).unwrap();
```

PR #211, opened 2021-04-09
quickstart: refer to re-exported object crate instead

tmplt:
#106 is hit if object and addr2line are added to Cargo.toml where object version does not match the version addr2line depends on. addr2line re-exports the correct version, which is now referred to instead.

Unfortunately, there doesn't seem to be a way to automagically link to the correct doc version for object in the quickstart.


philipc (approved):
Good idea, thanks! Ideally this crate should be using the latest `object` version anyway, so this doc version is fine. (I'm waiting to finish one more feature before doing a release.)

PR #213, opened 2021-04-21
Switch addr2line to use EndianReader.

### LINKED ISSUE #212: Doesn't seem to live up to performance claims?
I have a large binary (~256 MB before strip, 209 MB total) & I did a simple comparison, calling addr2line in a loop 20 times (providing 16 addresses from a real call stack) & comparing peak RSS & the total time taken. The command line invoked was
`<addr2line exe> --pretty-print -C -f -i  --exe <path to exe> <addresses>` (all output was piped to `/dev/null`).

The results are:
binutils: ~4s, 236 MB RSS
this crate: ~3s, 384 MB RSS
llvm-symbolizer: ~3s 138 MB RSS
elfutils: ~1s 53 MB RSS

I did notice that elfutils for some reason didn't manage to report the location of the function nor the function arguments so it's possible there's some bug or other issue with elfutils invalidating that result.

Regardless, it looks like llvm-symbolizer does use almost 1/3 the RAM for the same runtime performance & even binutils uses less RAM (albeit @ 33% slower runtime).

vlovich:
Avoid copying the debug into RAM & instead use borrows to utilize the 0-copy facility of gimli.
Resolves #212

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -4,15 +4,18 @@ extern crate fallible_iterator;
 extern crate gimli;
 extern crate memmap;
 extern crate object;
+extern crate typed_arena;
 
 use std::borrow::Cow;
 use std::fs::File;
 use std::io::{BufRead, Lines, StdinLock, Write};
 use std::path::Path;
+use std::result;
 
 use clap::{App, Arg, Values};
 use fallible_iterator::FallibleIterator;
-use object::Object;
+use object::{Object, ObjectSection};
+use typed_arena::Arena;
 
 use addr2line::{Context, Location};
 
@@ -35,7 +38,7 @@ impl<'a> Iterator for Addrs<'a> {
     fn next(&mut self) -> Option<u64> {
         let text = match *self {
             Addrs::Args(ref mut vals) => vals.next().map(Cow::from),
-            Addrs::Stdin(ref mut lines) => lines.next().map(Result::unwrap).map(Cow::from),
+            Addrs::Stdin(ref mut lines) => lines.next().map(result::Result::unwrap).map(Cow::from),
         };
         text.as_ref()
             .map(Cow::as_ref)
@@ -72,6 +75,23 @@ fn print_function(name: &str, language: Option<gimli::DwLang>, demangle: bool) {
         print!("{}", addr2line::demangle_auto(Cow::from(name), language));
     } else {
         print!("{}", name);
+    }
+}
+
+fn load_file_section<'input, 'arena, Endian: gimli::Endianity>(
+    id: gimli::SectionId,
+    file: &object::File<'input>,
+    endian: Endian,
+    arena_data: &'arena Arena<Cow<'input, [u8]>>,
+) -> Result<gimli::EndianSlice<'arena, Endian>, ()> {
+    // TODO: Unify with dwarfdump.rs in gimli.
+    let name = id.name();
+    match file.section_by_name(name) {
+        Some(section) => match section.uncompressed_data().unwrap() {
+            Cow::Borrowed(b) => Ok(gimli::EndianSlice::new(b, endian)),
+            Cow::Owned(b) => Ok(gimli::EndianSlice::new(arena_data.alloc(b.into()), endian)),
+        },
+        None => Ok(gimli::EndianSlice::new(&[][..], endian)),
     }
 }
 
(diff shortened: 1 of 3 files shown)


vlovich:
Hey @philipc, this diff on my binary is crashing at `thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', examples/addr2line.rs:99:46`. The section name is `.debug_loclists` (prior to that it successfully loads `.debug_loc`). I was wondering if you had any insights on where I have a mistake in this PR.

examples/addr2line.rs
```diff
@@ -75,6 +87,23 @@ fn print_function(name: &str, language: Option<gimli::DwLang>, demangle: bool) {
...
 
+fn load_file_section<'input, 'arena, Endian: gimli::Endianity>(
+    id: gimli::SectionId,
+    file: &object::File<'input>,
+    endian: Endian,
+    arena_data: &'arena Arena<Cow<'input, [u8]>>
+) -> Result<gimli::EndianSlice<'arena, Endian>> {
+    // TODO: Unify with dwarfdump.rs in gimli.
+    let name = id.name();
+    let section = file.section_by_name(name).unwrap();
```
philipc, line 98: Don't use `unwrap`, you need to handle the case where the section doesn't exist by returning an empty slice.

vlovich:
Sweet that worked! I can confirm memory usage drops 3x & is now better than llvm-symbolizer. Runtime is similarly ~3x faster.

vlovich:
Hmm... some weird test failures I'm unable to reproduce locally (running Arch)

philipc:
CI works. Probably it's the output equivalence test, which depends on the binutils version, since we are comparing our output with it.

examples/addr2line.rs
```diff
@@ -4,18 +4,30 @@ extern crate fallible_iterator;
...
 use clap::{App, Arg, Values};
 use fallible_iterator::FallibleIterator;
-use object::Object;
+use object::{Object, ObjectSection};
+use typed_arena::Arena;
 
 use addr2line::{Context, Location};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Error {
```
philipc, line 23: Let's leave this out for now. Better error handling would be fine, but you're not actually creating any instances of this yet (`load_file_section` always returns `Ok`).

examples/addr2line.rs
```diff
@@ -148,6 +177,9 @@ fn main() {
         )
         .get_matches();
 
+    let endian = gimli::NativeEndian;
```
philipc, line 180: Please use `RunTimeEndian`.

examples/addr2line.rs
```diff
@@ -169,9 +204,18 @@ fn main() {
...
+        } else {
+            Ok(gimli::EndianSlice::new(&[][..], endian))
+        }
+    };
 
     let symbols = object.symbol_map();
-    let ctx = Context::new_with_sup(object, sup_object.as_ref()).unwrap();
+    let dwarf = gimli::Dwarf::load(&mut load_section, &mut load_sup_section).unwrap();
+
+    let ctx = Context::from_dwarf(dwarf).unwrap();
```
philipc, line 218: Please use `from_dwarf_with_sup`.

philipc (changes requested):

examples/addr2line.rs
vlovich: I saw this & this part of the API wasn't clear to me (didn't seem to gel in a coherent way). Specifically, `Dwarf::Load` takes in a non-optional `sup` itself. If I did a `Dwarf::load` on the sup, what would be the sup I provide there? Is the sup section loader in `Dwarf::load` something else?

examples/addr2line.rs
vlovich: That's true, but `load_file_section` needs to return a Result. What should the error type be?

philipc: It can be `()`.

examples/addr2line.rs
philipc: Yeah this part of the API isn't great, and I've been playing with a different way of doing it, but I'm still not sure.

For now, you need to call `Dwarf::load` twice, the same as what `Context::new_with_sup` does.
- call Dwarf::load with `load_section` and `load_sup_section`. For this call, only `.debug_str` will be loaded from the supplementary object file.
- call Dwarf::load with `load_sup_section` and a closure that always returns empty sections.

This does mean that `.debug_str` will be loaded twice from the supplementary object file.

philipc (approved):
Thank you!

Issue #10, opened 2016-12-05
lib.rs is becoming monolithic, break it up into smaller files


jonhoo:
Currently, `src/lib.rs` is becoming quite large. It would be better if `Unit` was factored out into its own module. `Mapping::locate` could also use some modularization love.

PR #222, opened 2021-06-20
break up lib.rs into module (WIP)

FlyingCanoe:
I move `ResUnit` and function related struct to there own module.  I think more could be move to new module.

more precisely, I was thinking to create a location module, but I am a bit unsure as to what should be included in that module.

do you think that frame related struct should go in the location module or should it go in a separate module. It could also go in the function module.

I also think the module I have created could be more cleanly separated.

I would appreciate some feedback.


philipc:
Thanks for the PR! Are you doing this only for issue #10? I'll consider this some more, but I'm unsure whether we actually want to do this. For example, `ResUnit` is closely tied to the functionality of `ResDwarf` and `Context`, so personally I think it's easier to maintain these if they are in the same file.

FlyingCanoe:
I am doing this for the issue #10 and to improve my understanding of how the different piece fit together. 
since my understanding of this codebase is still limited it possible for my suggestion for the module boundary to be sub optimal.
I probably need to play with it a bit more.

now that you suggest it, it seam like it would work better if `ResUnit`, `ResDwarf` and `Context` were in the same module

philipc:
Note that #10 is an old issue, a lot has changed since then, and I'm not sure we even want to do it (one reason not to is that it makes it harder to follow the git history). Let's start with a small step to see how it goes, rather than trying to do everything at once. I think that moving the function parsing to a separate module is probably ok (and maybe a later PR could move line parsing to a separate module).

src/function.rs
```diff
@@ -0,0 +1,465 @@
+use core::{cmp::Ordering, iter};
+
+use alloc::{borrow::Cow, boxed::Box};
+
+use crate::{Error, RangeAttributes, ResDwarf, demangle_auto, lazy::LazyCell, maybe_small::{self, Vec}, name_attr};
```
philipc, line 5: Please don't use nested imports because they make it too hard to see the module dependencies. (eg lazy::LazyCell should be on a separate line).

philipc: `name_attr` should be moved to this module too.

FlyingCanoe: if we move `name_attr`, we should also move `name_entry` as it is only use in `name_attr`

FlyingCanoe:
I was overgenerous with the `pub(crate)` marking.
The last commit reduce the visibility of a handful of function.

FlyingCanoe:
I was overgenerous with the pub(crate) marking.
The last two commit reduce the visibility of a handful of item.

src/function.rs
```diff
@@ -0,0 +1,563 @@
...
+
+use alloc::{borrow::Cow, boxed::Box};
+
+use crate::demangle_auto;
+use crate::lazy::LazyCell;
+use crate::maybe_small;
+use crate::maybe_small::Vec;
+use crate::Error;
+use crate::RangeAttributes;
+use crate::ResDwarf;
```
philipc, line 11: Sorry, by nested import I meant using `::` within the `{}`. So please change to:
```
use alloc::borrow::Cow;
use alloc::boxed::Box;
use alloc::vec::Vec;
use core::cmp::Ordering;
use core::iter;

use crate::{demangle_auto, Error, RangeAttributes, ResDwarf};
use crate::lazy::LazyCell;
use crate::maybe_small;
```

(For this specific case, nested imports wouldn't be bad, but in general I find them harder to read.)

src/function.rs
```diff
@@ -0,0 +1,563 @@
...
+            dwarf,
+            inlined_functions,
+            inlined_addresses,
+            inlined_depth + 1,
+        )
+    }
+}
+
+/// A function name.
+pub struct FunctionName<R: gimli::Reader> {
```
philipc, line 454: This is more related to `Frame` than the rest of the module, so let's leave it in `lib.rs`.

Also FYI, this is a public struct, so if we were to move it into this module, we would need to add `pub use crate::function::FunctionName` to `lib.rs`.

src/lib.rs
```diff
@@ -53,6 +53,10 @@ use core::mem;
 use core::num::NonZeroU64;
 use core::u64;
 
+use crate::function::Function;
+use crate::function::FunctionName;
+use crate::function::Functions;
+use crate::function::InlinedFunction;
```
philipc, line 59:
```suggestion
use crate::function::{Function, FunctionName, Functions, InlinedFunction};
```

philipc (changes requested):

philipc:
The macOS failure may be due to https://github.com/rust-lang/cargo/pull/9298, because the test requires the dsym.

philipc (approved):
Thanks! If you can rebase on `master` then CI should pass.

FlyingCanoe:
I have rebase my PR. unless I have miss something, it should be ready to merge.

PR #224, opened 2021-06-23
Update "object" dependency

olivierlemasle:


philipc:
Thanks, but since this is a breaking change, I won't merge this yet. Also FYI, I do update dependencies when doing releases, and I assume that updating dependencies won't have any benefit for you until there is a release.

PR #225, opened 2021-07-02
Expose endian_reader as a separate feature.

gz:
Allows to build a Context with EndianRcSlice in no-std.
endian_reader is enabled by default on std but can't be enabled in no-std -- so building a context with EndianRcSlice results in the compiler unable to find it because it's not in the addr2line gimli version.

Maybe there is a better way to resolve this with cargo, but I'm not familiar with anything that would let me do this without patching addr2line.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,8 +54,9 @@ split-debuginfo = 'packed'
 
 [features]
 default = ["rustc-demangle", "cpp_demangle", "std-object", "fallible-iterator", "smallvec"]
+endian_reader = ["gimli/endian-reader"]
 std = ["gimli/std"]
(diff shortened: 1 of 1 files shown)


philipc:
You can add `gimli` to your own `Cargo.toml` and enable features there. It means you need to ensure you're using the same `gimli` version as `addr2line`, but that's not hard to do. I think it is better to do it this way, because an `endian_reader` feature does nothing for `addr2line` itself.

gz:
Hi, that probably makes more sense. I also thought it would work this way but got confused by the compiler errors there for a moment and couldn't figure out to get it to work but turns out I missed another dependency for addr2line form another Cargo.toml that I forgot to update as well.

Anyways will close this, thanks for the help, sorry for the noise.

PR #239, opened 2021-11-09
Dedupe comp-dir/directory in line-table paths

### LINKED ISSUE #238: Duplicated `comp-dir`
When building a simple rust program with `--remap-path-prefix $PWD=root-comp-dir`, that results in the following DWARF:

```
# CU (llvm-dwarfdump output):
0x0000000b:   DW_TAG_compile_unit
                DW_AT_producer	("clang LLVM (rustc version 1.56.1 (59eed8a2a 2021-11-01))")
                DW_AT_language	(DW_LANG_Rust)
                DW_AT_name	("simple.rs")
                DW_AT_stmt_list	(0x00000000)
                DW_AT_comp_dir	("root-comp-dir/")
                DW_AT_GNU_pubnames	(true)
                DW_AT_low_pc	(0x00000000000010f0)
                DW_AT_high_pc	(0x00000000000010f9)
0x0000002f:     DW_TAG_subprogram
                  DW_AT_low_pc	(0x00000000000010f0)
                  DW_AT_high_pc	(0x00000000000010f9)
                  DW_AT_frame_base	(DW_OP_reg7 RSP)
                  DW_AT_name	("simple_fn")
                  DW_AT_decl_file	("root-comp-dir/simple.rs")
                  DW_AT_decl_line	(2)
                  DW_AT_external	(true)

# Line Table:

Line table prologue:
          format: DWARF32
         version: 4
file_names[  1]:
           name: "simple.rs"
      dir_index: 0
```

The above example is using `llvm-dwarfdump`, which seems to get this right.

However, `addr2line` duplicates the `comp-dir`:

```
> /usr/bin/addr2line -i -e simple.debug 0x10f2
root-comp-dir//simple.rs:5
> ~/.cargo/bin/addr2line -i -e simple.debug 0x10f2
root-comp-dir/root-comp-dir/simple.rs:5
```

My system `addr2line` does not duplicate the comp-dir, but has problems when joining the paths it seems.

I think the root cause is that `LineProgramHeader::directory` is defined in such a way that it returns the `comp_dir` for index 0, but that is then unconditionally concatenated to the comp dir, similar to here:

https://github.com/gimli-rs/gimli/blob/fdd7ee4ed4c58bee9b79ffdc895488b40a87c899/examples/simple_line.rs#L71-L73

See: https://docs.rs/gimli/0.26.1/gimli/read/struct.LineProgramHeader.html#method.directory

I can get a PR up to fix this next week.

Swatinem:
fixes #238

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -701,14 +701,17 @@ impl<R: gimli::Reader> ResUnit<R> {
             String::new()
         };
 
-        if let Some(directory) = file.directory(header) {
-            path_push(
-                &mut path,
-                sections
-                    .attr_string(&self.dw_unit, directory)?
-                    .to_string_lossy()?
-                    .as_ref(),
-            );
+        // The directory index 0 is defined to correspond to the compilation unit directory.
+        if file.directory_index() != 0 {
+            if let Some(directory) = file.directory(header) {
+                path_push(
+                    &mut path,
+                    sections
+                        .attr_string(&self.dw_unit, directory)?
+                        .to_string_lossy()?
+                        .as_ref(),
+                );
+            }
         }
 
         path_push(


src/lib.rs
```diff
@@ -702,13 +702,11 @@ impl<R: gimli::Reader> ResUnit<R> {
...
-            path_push(
-                &mut path,
-                sections
-                    .attr_string(&self.dw_unit, directory)?
-                    .to_string_lossy()?
-                    .as_ref(),
-            );
+            let directory = sections.attr_string(&self.dw_unit, directory)?;
+            let directory = directory.to_string_lossy()?;
+            if directory != path {
```
philipc, line 707: This doesn't seem exactly right. While unlikely, it would be possible for these to be equal but still need to be relative.

Can we add a special case for directory index 0 instead?

Swatinem: sure!

src/lib.rs
```diff
@@ -701,14 +701,18 @@ impl<R: gimli::Reader> ResUnit<R> {
...
-            path_push(
-                &mut path,
-                sections
-                    .attr_string(&self.dw_unit, directory)?
-                    .to_string_lossy()?
-                    .as_ref(),
-            );
+        // In DWARF version <= 4, the index 0 is defined to correspond to the
+        // compilation unit directory.
+        if !(header.version() <= 4 && file.directory_index() == 0) {
```
philipc, line 706: DWARF 5 says: "The first entry is the current directory of the compilation. Each additional path entry is either a full path name or is relative to the current directory of the compilation. The line number program assigns a number (index) to each of the directory entries in order, beginning with 0."

So I don't think we should restrict this to only DWARF version <= 4.

Swatinem: Interesting. Well the `LineProgramHeader` uses `<= 4` as a cutoff point, both for directories and file names further down below:

https://github.com/gimli-rs/gimli/blob/fdd7ee4ed4c58bee9b79ffdc895488b40a87c899/src/read/line.rs#L1168-L1190

Should I update all of that and rather continue with https://github.com/gimli-rs/gimli/pull/604 first?

philipc: The change in version 5 (and the reason for <=4 as a cutoff in that link) is that previously the compilation directory entry was not included in `.debug_line`, but a directory index of 0 still means the same for all versions.

There is a separate issue where earlier in this function we probably should be using `header.directory(0)` instead of `self.dw_unit.comp_dir`, but in theory they should be equivalent.

philipc: (So the code you linked to in `line.rs` doesn't need fixing, it is correct; all it is doing is handling the case where the compilation directory is not present in `self.include_directories` so it has to use `self.comp_dir` instead.)

Swatinem: Thanks for the explanation!

philipc (approved):
Thanks!

PR #242, opened 2022-03-13
Update clap to v3.1

LovecraftianHorror:
This updates the `addr2line` example to `clap` v3.1.6. `clap` does have an MSRV of 1.54.0 on any v3 version, but this is unlikely to matter much in this case since it is just used as a dev dependency

The changes were pretty straightforward (and made the example more succinct), and everything seems to match the old behavior

_Just a basic test_

```
     Running `target/debug/examples/addr2line -e /tmp/a.out -f 0x1139`
main
/tmp/test.c:3
```

_Old `--help`_

```
hardliner 0.1
A fast addr2line clone

USAGE:
    addr2line [FLAGS] [OPTIONS] --exe <filename> [addrs]...

FLAGS:
    -a, --addresses       Display the address before the function name, file and line number information.
    -s, --basenames       Display only the base of each file name.
    -C, --demangle        Demangle function names. Specifying a specific demangling style (like GNU addr2line) is not
                          supported. (TODO)
    -f, --functions       Display function names as well as file and line number information.
    -h, --help            Prints help information
    -i, --inlines         If the address belongs to a function that was inlined, the source information for all
                          enclosing scopes back to the first non-inlined function will also be printed.
        --llvm            Display output in the same format as llvm-symbolizer.
    -p, --pretty-print    Make the output more human friendly: each location are printed on one line.
    -V, --version         Prints version information

OPTIONS:
    -e, --exe <filename>    Specify the name of the executable for which addresses should be translated.
        --sup <filename>    Path to supplementary object file.

ARGS:
    <addrs>...    Addresses to use instead of reading from stdin.
```

_New `--help`_

```
hardliner 0.1
A fast addr2line clone

USAGE:
    hardliner [OPTIONS] --exe <FILENAME> [ADDRS]...

ARGS:
    <ADDRS>...    Addresses to use instead of reading from stdin

OPTIONS:
    -a, --addresses         Display the address before the function name, file and line number
                            information
    -C, --demangle          Demangle function names. Specifying a specific demangling style (like
                            GNU addr2line) is not supported. (TODO)
    -e, --exe <FILENAME>    Specify the name of the executable for which addresses should be
                            translated
    -f, --functions         Display function names as well as file and line number information
    -h, --help              Print help information
    -i, --inlines           If the address belongs to a function that was inlined, the source
                            information for all enclosing scopes back to the first non-inlined
                            function will be printed
        --llvm              Display output in the same format as llvm-symbolizer
    -p, --pretty-print      Make the output more human friendly: each location is printed on one
                            line
    -s, --basenames         Display only the base of each file name
        --sup <FILENAME>    Path to supplementary object file
    -V, --version           Print version information
```

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -96,6 +96,7 @@ fn load_file_section<'input, 'arena, Endian: gimli::Endianity>(
 }
 
 #[derive(Parser)]
+#[clap(bin_name = "hardliner")]
 #[clap(version = "0.1")]
 #[clap(about = "A fast addr2line clone")]
 struct Opts {
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -9,10 +9,9 @@ extern crate typed_arena;
 use std::borrow::Cow;
 use std::fs::File;
 use std::io::{BufRead, Lines, StdinLock, Write};
-use std::path::{Path, PathBuf};
-use std::vec::IntoIter;
+use std::path::Path;
 
-use clap::Parser;
+use clap::{Arg, Command, Values};
 use fallible_iterator::FallibleIterator;
 use object::{Object, ObjectSection};
 use typed_arena::Arena;
@@ -28,7 +27,7 @@ fn parse_uint_from_hex_string(string: &str) -> u64 {
 }
 
 enum Addrs<'a> {
-    Args(IntoIter<String>),
+    Args(Values<'a>),
     Stdin(Lines<StdinLock<'a>>),
 }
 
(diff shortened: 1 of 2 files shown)


LovecraftianHorror:
It looks like the CI failure is from unrelated changes with a more recent rust update since stable passed, but beta and nightly both failed

examples/addr2line.rs
```diff
@@ -94,90 +95,60 @@ fn load_file_section<'input, 'arena, Endian: gimli::Endianity>(
     }
 }
 
+#[derive(Parser)]
+#[clap(bin_name = "hardliner", name = "hardliner")]
```
philipc, line 99: The binary that rust builds is named `addr2line`, so this isn't right. Let's remove the `hardliner` name completely, since it's an old name that is no longer used anywhere.

philipc (changes requested):
Thanks for the PR, but since we are already using the builder API, I prefer to stay with that instead of changing to the derive API. If you wanted to make the builder API slightly more compact you could change it to use `Command::args`.

LovecraftianHorror:
Thanks for the feedback! 039be6e56befa301d2ccebec1dd19ef3d40d880a switched back to the builder API, switched the name fully over to `addr2line` (and changed the about to make it's distinction clear), switched to using `.args`, and normalized the help help lines for all the args.

Let me know if there is anything you want changed

philipc (approved):

PR #244, opened 2022-04-27
cli: Fixes and refactoring for perf support

ishitatsuyuki:
#### Background

The _perf_ utility uses either _libbfd_ or _addr2line_ to obtain information about inlined symbols (the record-time unwinding itself is handled by _libunwind_). Both option uses _libbfd_ under the hood, and it seems that _libbfd_ lacks the functionality of properly constructing a binary search index before doing the mass lookups. As a result, it runs like hot garbage for a lot of debug binaries.

The option to use an out-of-process _addr2line_ was originally added for Debian linking license reasons, but it turns out that with a little tweaking we can drop in a different addr2line implementation as well. With these modifications and Gimli's addr2line dropped into /usr/local/bin, `perf report` runs magnitudes faster (~100x? didn't measure) and there doesn't appear to be major issues.

**Important**: If you want to actually try, you need to (re)build perf with the `NO_LIBBFD` option passed to make. Probably only Debian has this by default.

#### Description

perf has a few assumptions about how addr2line behaves, and here we align two behavior differences with the GNU implementation:
1. We just give a "no match" when we can't parse the input address.
2. We output `??:0` instead of `??:?` for "no match".

There are also a few refactoring so that the output code can be streamlined.

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -45,8 +45,8 @@ impl<'a> Iterator for Addrs<'a> {
     }
 }
 
-fn print_loc(loc: &Option<Location>, basenames: bool, llvm: bool) {
-    if let Some(ref loc) = *loc {
+fn print_loc(loc: Option<&Location>, basenames: bool, llvm: bool) {
+    if let Some(ref loc) = loc {
         let file = loc.file.as_ref().unwrap();
         let path = if basenames {
             Path::new(Path::new(file).file_name().unwrap())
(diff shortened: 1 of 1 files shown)


mstange:
This sounds excellent!

Does `perf` use one long-running `addr2line` process these days, or does it relaunch `addr2line` for every address? If it relaunches for every address, then you're not benefiting from the addr2line-internal caching and could go even faster. (I saw [this post about using a long-running process](https://eighty-twenty.org/2021/09/09/perf-addr2line-speed-improvement) but I do not know whether this patch has made it into perf.)

ishitatsuyuki:
I’m running a build from 5.17 source tree and it’s using a persistent process.

philipc (changes requested):
These changes are great. Please rebase onto `master` for CI.

ishitatsuyuki:
Rebased. Please approve CI as I'm a first-time contributor.

philipc (approved):

PR #249, opened 2022-05-22
Improve compatibility with .so files produced by perf inject --jit

### LINKED ISSUE #248: No line information found in jitted-1234-56.so files generated by perf inject --jit
No filename and line number is found in [this jitted-103168-33.so](https://storage.googleapis.com/profiler-get-symbols-fixtures/jitted-103168-33.so):

```
% addr2line -pCfi 0x82 --exe jitted-103168-33.so  
IC: GlobalNameValue at ??:?
```

This .so has a rather minimal set of dwarf information, generated by [perf/util/genelf_debug.c](https://elixir.bootlin.com/linux/v5.17.9/source/tools/perf/util/genelf_debug.c#L434).
The single unit has no low pc / high pc information, so addr2line collects no unit ranges, so all address lookups fail immediately because no unit is found.

These .so files are produced by [`perf inject --jit`](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/perf-inject.txt) after using perf to profile applications with [JITDUMP](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jitdump-specification.txt) support.

mstange:
Fixes #248.

In addition to the missing unit ranges, there was a bug where we weren't able to find the file name that was added during line rows iteration, and we were creating an absolute path instead of a relative path.

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -47,13 +47,16 @@ impl<'a> Iterator for Addrs<'a> {
 
 fn print_loc(loc: Option<&Location>, basenames: bool, llvm: bool) {
     if let Some(ref loc) = loc {
-        let file = loc.file.as_ref().unwrap();
-        let path = if basenames {
-            Path::new(Path::new(file).file_name().unwrap())
+        if let Some(ref file) = loc.file.as_ref() {
+            let path = if basenames {
+                Path::new(Path::new(file).file_name().unwrap())
+            } else {
+                Path::new(file)
+            };
+            print!("{}:", path.display());
         } else {
-            Path::new(file)
-        };
-        print!("{}:", path.display());
+            print!("??:");
+        }
         if llvm {
             print!("{}:{}", loc.line.unwrap_or(0), loc.column.unwrap_or(0));
         } else if let Some(line) = loc.line {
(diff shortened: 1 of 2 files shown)


mstange:
I considered adding a test for this by dropping the 1KB test file into the fixtures directory, but I couldn't find a good place to add the test without creating a bunch of new infrastructure. Most of the tests are rather self-referential. Only the parse.rs tests make use of a frozen file (`fixtures/addr2line-release`), and those don't check for correctness, only for non-panic.

mstange:
I could rename `correctness.rs` to `correctness_self.rs` and add a new `correctness_frozen.rs`. How does that sound? Or maybe `_snapshot` or `_fixtures`.

philipc:
The current `correctness.rs` is basically checking that we can still handle whatever rustc generates.

We do need to add regression tests though (both for this and ideally for all the previous bugs that have been fixed). Perhaps `regression.rs` would be suitable for that, but I don't have any preference. If we were to regression test properly, we could potentially end up with a largish number of binary files. I'm not sure that we want to put them in this repository. For `object`, I created the `object-testfiles` repository (which we could reuse); I'm not sure that's a good solution either, since having to update both of them is cumbersome. Open to any suggestions here.

philipc (approved):
Thanks!

PR #255, opened 2022-08-25
Replace clap with bpaf

pacak:
So I did a thing and tried to replace clap's build API with bpaf's one.

Summary changes:

- Changed to use `PathBuf` instead of going though `String`.

   You can check the difference by running `cargo run --release --example addr2line -- --exe $( echo "\xe9" )`, older version would die complaining about non-utf8 symbols, new version would complain about missing file. This can be somewhat important on windows or with strange locales on unix.

- Collected options in a struct and made them strictly typed.

   Previous version used strings to refer to names and while everything seemed worked fine it is way too easy to make a typo while adding a new field defining it as `"help"` and asking for it later as `"hepl"`. Similar with types. Now`pretty` is bool, not something you need to ask for separately.

- I changed a bit how addresses are processed, output seems to match, but let me know - I can replicate the old behavior exactly.

- Version flag now works and takes version from Cargo.toml, current version is 0.18.0

- Clean release compilation time for the whole project went down from 14 to 10 seconds on my machine.

- Executable size went down from 26Mb to 14Mb

- `--help` looks a bit different and no longer green.
```console
% cargo run --release --example addr2line -- --help
A fast addr2line Rust port

Usage: [-a] [-f] -e filename [--sup filename] [-p] [-C] [-s] [-i] [-L] <addrs>...

Available positional items:
    <addrs>  Addresses to use instead of reading from stdin.

Available options:
    -a, --addresses       Display the address before the function name, file and line number information.
    -f, --functions       Display function names as well as file and line number information.
    -e, --exe <filename>  Specify the name of the executable for which addresses should be translated.
        --sup <filename>  Path to supplementary object file.
    -p, --pretty-print    Make the output more human friendly: each location are printed on one line.
    -C, --demangle        Demangle function names.
                          Specifying a specific demangling style (like GNU addr2line) is not supported. (TODO)
    -s, --basenames       Display only the base of each file name.
    -i, --inlines         If the address belongs to a function that was inlined, the source information for
                          all enclosing scopes back to the first non-inlined function will also be printed.
    -L, --llvm            Display output in the same format as llvm-symbolizer.
    -h, --help            Prints help information
    -V, --version         Prints version information
```

PS: https://crates.io/crates/bpaf - something I made, this code uses combinatoric API, you can find a tutorial here: https://docs.rs/bpaf/0.5.3/bpaf/_derive_tutorial/index.html

I'm happy to answer any questions you might have.

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -1,5 +1,5 @@
 extern crate addr2line;
-extern crate clap;
+extern crate bpaf;
 extern crate fallible_iterator;
 extern crate gimli;
 extern crate memmap2;
@@ -8,45 +8,25 @@ extern crate typed_arena;
 
 use std::borrow::Cow;
 use std::fs::File;
-use std::io::{BufRead, Lines, StdinLock, Write};
-use std::path::Path;
+use std::io::{BufRead, Write};
+use std::path::{Path, PathBuf};
 
-use clap::{Arg, Command, Values};
 use fallible_iterator::FallibleIterator;
 use object::{Object, ObjectSection, SymbolMap, SymbolMapName};
 use typed_arena::Arena;
 
 use addr2line::{Context, Location};
 
-fn parse_uint_from_hex_string(string: &str) -> Option<u64> {
+fn parse_uint_from_hex_string(string: &str) -> Result<u64, std::num::ParseIntError> {
     if string.len() > 2 && string.starts_with("0x") {
-        u64::from_str_radix(&string[2..], 16).ok()
+        u64::from_str_radix(&string[2..], 16)
     } else {
-        u64::from_str_radix(string, 16).ok()
-    }
-}
-
-enum Addrs<'a> {
-    Args(Values<'a>),
-    Stdin(Lines<StdinLock<'a>>),
-}
-
-impl<'a> Iterator for Addrs<'a> {
-    type Item = Option<u64>;
-
-    fn next(&mut self) -> Option<Option<u64>> {
-        let text = match *self {
-            Addrs::Args(ref mut vals) => vals.next().map(Cow::from),
-            Addrs::Stdin(ref mut lines) => lines.next().map(Result::unwrap).map(Cow::from),
-        };
-        text.as_ref()
-            .map(Cow::as_ref)
-            .map(parse_uint_from_hex_string)
+        u64::from_str_radix(string, 16)
     }
 }
 
 fn print_loc(loc: Option<&Location>, basenames: bool, llvm: bool) {
-    if let Some(ref loc) = loc {
+    if let Some(loc) = loc {
         if let Some(ref file) = loc.file.as_ref() {
             let path = if basenames {
                 Path::new(Path::new(file).file_name().unwrap())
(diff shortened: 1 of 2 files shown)


philipc:
Hi, thanks for the PR. The CI failure is probably unrelated to this PR.

Many of these changes certainly sound useful. However, I'm reluctant to switch this crate to use a dependency that appears to have few other users at this time. The example application is not an area where I want to be breaking new ground. Which of these changes would not be possible to achieve while still using `clap`? Is `bpaf` viable as a replacement for all users of `clap`?

pacak:
> However, I'm reluctant to switch this crate to use a dependency that appears to have few other users at this time.

Fair enough, I won't be objecting to closing this PR.

> Which of these changes would not be possible to achieve while still using clap?

From developer point of you to get typo safety you'll have to switch to derive version.  Big things you won't be able to achieve is compilation time and binary size.

> Is `bpaf` viable as a replacement for all users of `clap`?

Getting there, in fact this PR served it's purpose for this as well - I'm looking at real world examples. Currently released version lacks some flexibility in filename parsing - this was fixed in unpublished version, with that in place it should have feature parity, at least as far as long as we are comparing base crates. Planning to release a new version it once I finish implementing dynamic shell completion and we'll see what kind of feedback I'll get from clap maintainers. So maybe wait a bit and I'll post a link to the discussion once I have it.

pacak:
https://www.reddit.com/r/rust/comments/wvp9ao/yet_another_command_line_argument_parser_bpaf_052/ - discussion from the previous release and everything other than `-fbar` and colors can be achieved.

pacak:
Released a new version, here's a discussion:
https://www.reddit.com/r/rust/comments/x3z5xn/yet_another_command_line_argument_parser_bpaf_055/

But as I said earlier - no objection to closing this PR, it served it's purpose for me.

pacak:
Also added support for shell completion - because why not :)

philipc:
Thanks, I am interested in this. I'll review it later.

examples/addr2line.rs
```diff
@@ -203,85 +219,96 @@ fn main() {
...
 
-    for probe in addrs {
-        if print_addrs {
-            let addr = probe.unwrap_or(0);
-            if llvm {
+    for mprobe in addrs {
+        let probe = match mprobe {
+            Some(probe) => probe,
+            None => {
+                println!("??:0");
```
philipc, line 237: I think we should call `print_loc` here to get the `opts.llvm` handling

philipc (changes requested):

philipc:
> From developer point of you to get typo safety you'll have to switch to derive version. Big things you won't be able to achieve is compilation time and binary size.

I think this is a good enough reason to change. Thanks!

examples/addr2line.rs
pacak: Done.

pacak:
Would you be interested in similar changes to other projects in gimli-rs organization? `gimli` seems to be using `getopts` to handle them doing more stringly typed things. Might take a week or two though.

philipc:
Thanks, I'll merge this once I have CI working again.

> Would you be interested in similar changes to other projects in gimli-rs organization? gimli seems to be using getopts to handle them doing more stringly typed things. Might take a week or two though.

Let's leave that for now. I think `gimli` is simple enough how it is. `ddbug` might benefit, but it's certainly not a priority.

philipc:
Sorry for taking so long to merge this. I've changed my mind and I'm going to stay with clap for now. I've incorporated some of the improvements from this PR in #266. Thank you very much for working on these.

A couple of notes about this PR:

> Executable size went down from 26Mb to 14Mb

This includes debuginfo, which isn't as relevant.

> I changed a bit how addresses are processed, output seems to match

This isn't matching for the case where the address is invalid. I think it's better to leave it how it was.

Issue #256, opened 2022-09-13
Tests are failing with nightly rust due to packed debuginfo


philipc:
Currently the tests require Linux builds to use `split-debuginfo=off` and macOS builds to use `split-debuginfo=packed`. On stable rust, we achieve this with `split-debuginfo=packed` in `Cargo.toml`. This works because this option is only supported for macOS. However, nightly rust has stabilized this option for Linux as well, and so now CI is failing.

I can't see how to specify platform specific options in profiles. For example, I would expect something like this to work, but it doesn't:
```
[target.'cfg(macos)'.profile.test]
split-debuginfo = 'packed'
```

@davidtwco Do you know how to achieve this? Is it intended that all platforms must use the same `split-debuginfo` settings?

bjorn3:
macOS defaults to packed for back compat reasons, so not setting split-debuginfo at all should work.

philipc:
Hmm, can't test that now, but the reason I had this in the first place is because it was defaulting to unpacked.

bjorn3:
Right, cargo changed the default in https://github.com/rust-lang/cargo/pull/9298. My bad.

philipc:
The relevant cargo issue is https://github.com/rust-lang/cargo/issues/4897

We can use environment variables to work around this in CI, but the cargo setting would still be preferred.

PR #257, opened 2022-10-26
Bump cpp_demangle to 0.4.

khuey:


Cargo.toml
```diff
@@ -50,7 +50,7 @@ split-debuginfo = 'packed'
 
 [features]
 default = ["rustc-demangle", "cpp_demangle", "std-object", "fallible-iterator", "smallvec"]
-std = ["gimli/std"]
+std = ["gimli/std", "cpp_demangle/std"]
```
philipc, line 53: This doesn't appear to be needed, and forces `cpp_demangle` enabled.

khuey: Hmm, it appears that `addr2line` just tosses errors from `cpp_demangle`, so yeah, that seems true.

khuey:
Poke.

philipc (approved):

philipc:
Looks fine. I still need to find a way to fix #256 before CI will pass. I normally do this sort of change as part of the release.

PR #263, opened 2023-02-03
Refactor split DWARF support

philipc:
See individual commits.

cc @khuey

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -10,7 +10,6 @@ use std::env;
 use std::fs::File;
 use std::path::{self, PathBuf};
 
-use addr2line::LookupResultExt;
 use object::{Object, ObjectSection, ObjectSymbol};
 
 fn release_fixture_path() -> PathBuf {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -737,31 +745,60 @@ impl<R: gimli::Reader> ParsedDwarf<R> {
...
-            DebugFile::Supplementary => self
-                .sup_units
-                .as_ref()
-                .ok_or(gimli::Error::NoEntryAtGivenOffset)?,
+    ) -> Result<(&gimli::Unit<R>, gimli::UnitOffset<R::Offset>), Error> {
+        let unit = match file {
+            DebugFile::Primary => {
+                match self
+                    .units
+                    .binary_search_by_key(&offset.0, |unit| unit.offset.0)
```
khuey, line 778: Duplicating the binary search logic strikes me as less clear than what was there before.

benches/bench.rs
```diff
@@ -10,7 +10,6 @@ use std::env;
 use std::fs::File;
 use std::path::{self, PathBuf};
 
-use addr2line::LookupResultExt;
```
khuey, line 13: You need to remove this from the example too. You didn't hit it because it's #[cfg(not(unix))]

khuey (approved):

src/lib.rs
philipc: True, but they are different types now so I didn't see how we could reuse it anymore.

PR #268, opened 2023-02-17
Update to Edition 2018

kornelski:
The crate isn't compatible with any 2015-only compiler, so it can use a more recent Rust edition.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,7 +36,6 @@ extern crate alloc;
 extern crate cpp_demangle;
 #[cfg(feature = "fallible-iterator")]
 pub extern crate fallible_iterator;
-pub extern crate gimli;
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -36,7 +36,6 @@ extern crate alloc;
 extern crate cpp_demangle;
 #[cfg(feature = "fallible-iterator")]
 pub extern crate fallible_iterator;
```
philipc, line 38: Is there a reason the feature gated crates can't be removed too?

kornelski: These could be removed too, just wasn't handled by `cargo fix`

philipc: After a second look, the public crates need to be kept. Maybe the removal of gimli below is a cargo fix bug.

PR #275, opened 2023-04-16
Set rust-version in Cargo.toml

atouchet:
Now that the MSRV is higher than 1.56 it should be safe to set it in Cargo.toml.


philipc (approved):
Thanks!

PR #277, opened 2023-07-04
feat: modify Cargo.toml to port to shyper

hky1999:


philipc:
What is shyper? Why does it need this? Can you add a dependency on gimli in your crate, and enable the feature there instead? I'm reluctant to add features that don't directly affect the functionality of this crate.

hky1999:
Sorry to bother you, I forked my own repo, but I made a mistake when creating a MR.
Sorry, best wishes

PR #278, opened 2023-07-18
Store boxed slices instead of `Vec` objects in `Context`

danielocfb:
We never resize the unit ranges, units, or sup units that we parsed -- they stay what they were when we instantiated the Context object. As such, we don't need to store `Vec` objects, which include a capacity. Do what we do for the `Functions` type and just store boxed slices. Doing so cuts a machine word of each, but also trims heap allocations to the minimal size required in the process.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,9 +183,9 @@ impl<L: LookupContinuation> LookupResult<L> {
 /// when performing lookups for many addresses in the same executable.
 pub struct Context<R: gimli::Reader> {
     sections: Arc<gimli::Dwarf<R>>,
-    unit_ranges: Vec<UnitRange>,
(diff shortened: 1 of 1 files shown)


bjorn3:
If capacity and size don't match trimming is expensive as it requires a copy of all data in many cases.

philipc:
None of the benchmarks show any significant difference, so seems fine to me. Is this patch due to something you noticed in use, or only from code review?

danielocfb:
> Is this patch due to something you noticed in use, or only from code review?

It's from code review.

If you don't like it then feel free to close, no hard feelings. But to me trimming in most places but not in others seems bogus, given that we are not modifying these structures afterwards by design.

PR #281, opened 2023-08-30
Remove scripts from package

Jake-Shadle:
There is no reason to include the shell/R scripts in the final package.


philipc (approved):
Thanks!

PR #309, opened 2024-06-30
Fix clippy warnings from 1.79.0 and 1.80.0:

marxin:


tests/output_equivalence.rs
```diff
@@ -119,7 +119,7 @@ fn make_tests() -> Vec<Trial> {
         .map(|param| {
             Trial::test(
                 format!("addr2line {}", param.as_ref().map_or("", String::as_str)),
-                move || run_test(param.as_ref().map(String::as_str)),
+                move || run_test(param.as_deref()),
```
philipc, line 122: Can you also change the previous line to use `as_deref`?

marxin: Sure, done!

marxin (commented):
@philipc Btw. what's the reason this particular test (`output_equivalence`) is disabled on Linux? It runs fine if I see correctlly.

philipc (approved):

PR #312, opened 2024-07-24
Also exclude rustfmt.toml from published crate

michel-slm:
This is not needed by non-developers, and we've been patching our downstream Cargo.toml in Fedora to exclude it.


philipc (changes requested):
It's an empty file. Can we just delete it instead?

Also, not for this PR, but it might be better to switch to using include.

PR #315, opened 2024-07-25
Add benchmark-addr2line.py script

marxin:
Add a benchmarking script that compares binutils, elfutils, LLVM, and this project implementation of `addr2line`.

The limitation of `cargo bench` is that it's designed for micro-benchmarking, where a benchmark is run ~100 times. Thus I came up with a separate Bash script that utilizes `hyperfine`.

Sample output:
```
Benchmark 1: gimli-addr2line libxul.so
  Time (mean ± σ):     115.5 ms ±   2.4 ms    [User: 86.1 ms, System: 29.3 ms]
  Range (min … max):   110.9 ms … 120.7 ms    25 runs
 
Benchmark 2: binutils-addr2line libxul.so
  Time (mean ± σ):      1.196 s ±  0.017 s    [User: 0.962 s, System: 0.233 s]
  Range (min … max):    1.184 s …  1.208 s    2 runs
 
Benchmark 3: llvm-addr2line libxul.so
  Time (mean ± σ):     493.1 ms ±   7.3 ms    [User: 445.9 ms, System: 47.1 ms]
  Range (min … max):   481.2 ms … 501.2 ms    6 runs
 
Benchmark 4: elfutils-addr2line libxul.so
  Time (abs ≡):         3.696 s               [User: 3.676 s, System: 0.018 s]
 
Relative speed comparison
        1.00          gimli-addr2line libxul.so
       10.35 ±  0.26  binutils-addr2line libxul.so
        4.27 ±  0.11  llvm-addr2line libxul.so
       31.99          elfutils-addr2line libxul.so
Benchmark 1: gimli-addr2line clangd-18
  Time (mean ± σ):     391.1 ms ±   4.2 ms    [User: 252.2 ms, System: 138.8 ms]
  Range (min … max):   383.9 ms … 395.7 ms    7 runs
 
Benchmark 2: binutils-addr2line clangd-18
  Time (abs ≡):         3.115 s               [User: 2.456 s, System: 0.658 s]
 
Benchmark 3: llvm-addr2line clangd-18
  Time (mean ± σ):     398.0 ms ±   3.7 ms    [User: 323.9 ms, System: 74.0 ms]
  Range (min … max):   393.8 ms … 405.4 ms    7 runs
 
Benchmark 4: elfutils-addr2line clangd-18
  Time (abs ≡):         2.393 s               [User: 2.359 s, System: 0.033 s]
 
Relative speed comparison
        1.00          gimli-addr2line clangd-18
        7.96          binutils-addr2line clangd-18
        1.02 ±  0.01  llvm-addr2line clangd-18
        6.12          elfutils-addr2line clangd-18
...
```


philipc (commented):
Can you add a script so that I can run this in a container on my dev system? For example, in the backtrace crate I can run [ci/debuglink-docker.sh](https://github.com/rust-lang/backtrace-rs/blob/master/ci/debuglink-docker.sh) to run the debuglink tests.

Did you consider using something like hyperfine or pyperf for the timing?

marxin:
All right. After a little investigation, I utilized hyperfine which ends up with a minimal shell script. What do you think about it?

When it comes to the Docker script: do you want a script that will build a Docker image (`docker build`) and run it then? Should it be the same script the GitHub CI should run?

philipc:
The hyperfine output isn't as pretty as the table you had, but I still think it's worth using it to get more reliable results.

Yes, I want a single command to run the benchmark on my non-openSUSE system, and it should avoid running the zypper commands every time. I'm not particularly familiar with using docker in github actions, but the current action spent a minute installing the packages, so seems like it might be a good idea to build an docker image for it too?

marxin:
> The hyperfine output isn't as pretty as the table you had, but I still think it's worth using it to get more reliable results.

Yeah, I think so. Moreover, one can use `--export-markdown` option to export the results and include them in `README.md` file.

> Yes, I want a single command to run the benchmark on my non-openSUSE system, and it should avoid running the zypper commands every time.

Sure. If we run `docker build && docker run` in the script, then the container building will be cached and you won't have to install all the packages every time.

> I'm not particularly familiar with using docker in github actions, but the current action spent a minute installing the packages, so seems like it might be a good idea to build an docker image for it too?

Me neither and I haven't found any easy solution for that. So do you want build a docker container that will be manually pushed to dockerhub and then used in the CI. What do you think?

philipc:
> Me neither and I haven't found any easy solution for that. So do you want build a docker container that will be manually pushed to dockerhub and then used in the CI. What do you think?

What you've got is fine. I don't think we want to push to dockerhub, it may be possible to [change the caching in the action to keep the layers](https://depot.dev/blog/docker-layer-caching-in-github-actions). I don't know. Only look into it more if you're keen.

marxin:
All right, so let's leave the caching for now. I've just updated the PR so that the requested wrapper script is introduced (and used by GitHub CI). Can you please review it?

philipc (changes requested):
This is looking good.

One last nit: can we move these into a subdirectory? e.g. scripts? And maybe a scripts/docker for the dockerfile, since it uses the subdirectory as context for the build, so best to restrict it to what is needed.

There'll need to be some cleanup of the existing scripts too, but that can another PR.

philipc:
Also this has already been useful in highlighting the need for #318

marxin:
> One last nit: can we move these into a subdirectory? e.g. scripts? And maybe a scripts/docker for the dockerfile, since it uses the subdirectory as context for the build, so best to restrict it to what is needed.

Sure, it should be done right now.

philipc (approved):
Thanks!

PR #318, opened 2024-07-28
Enable abbreviations cache for duplicates

philipc:
This speeds up the initial parsing of units when there are many small units that share abbreviations.

This was encountered for postgresql in opensuse/tumbleweed.

### DIFF
--- a/src/loader.rs
+++ b/src/loader.rs
@@ -209,6 +209,7 @@ impl<'a> LoaderInternal<'a> {
         if let Some(sup_object) = &sup_object {
             dwarf.load_sup(|id| load_section(Some(id.name()), sup_object, endian, arena_data))?;
         }
+        dwarf.populate_abbreviations_cache(gimli::AbbreviationsCacheStrategy::Duplicates);
 
         let ctx = Context::from_dwarf(dwarf)?;
 


marxin:
How much did it improve the numbers for gimli-addr2line for the particular benchmark?

philipc:
Roughly double. In the CI runs for #315, it was:
```
Relative speed comparison
        2.12 ±  0.02  gimli-addr2line postgres
        2.29 ±  0.04  binutils-addr2line postgres
        1.00          llvm-addr2line postgres
       24.41          elfutils-addr2line postgres
```
and after rebasing it improved to:
```
Relative speed comparison
        1.00          gimli-addr2line postgres
        2.71 ±  0.03  binutils-addr2line postgres
        1.20 ±  0.01  llvm-addr2line postgres
       29.51          elfutils-addr2line postgres
```

marxin:
Nice! I can reproduce it now (I forgot to run `cargo b --release` after I pulled the corresponding commit)! Do you have any other interesting binaries/shared libs we can add to the benchmark script?

philipc:
Probably not anything in particular that is interesting for performance reasons.

Possible things to look for are differences in things like frequency of inlined functions, use of sequences in the line table, use of shared abbreviations, size of compilation units. Some of those are going to depend on the compiler, compiler flags, or language.

What were the reasons for choosing the existing ones?

marxin:
> What were the reasons for choosing the existing ones?

Well, I basically took the biggest binaries I know of (Firefox, Clang) and then added what Mold uses for benchmarking: https://github.com/rui314/mold.

PR #325, opened 2024-08-05
Grouped bar chart graph

marxin:
The graph is based on the following example: https://matplotlib.org/stable/gallery/lines_bars_and_markers/barchart.html#sphx-glr-gallery-lines-bars-and-markers-barchart-py.

Apparently, we can use vector images in the README page, yay!


scripts/benchmark-plot.py
```diff
@@ -1,38 +1,50 @@
...
         results = json.load(f)["results"]
-    commands = [b["command"] for b in results]
-    data.append([b["mean"] for b in results])
+    if not commands:
+        # initialize the dictionary if empty
+        commands = {b["command"]: [] for b in results}
+    for result in results:
+        commands[result["command"]].append(
+            statistics.mean([result["mean"] for b in results])
+        )
```
philipc, line 28: Calculating the mean of a mean doesn't make sense to me, and nothing in this `for` loop needed to change anyway, so I prefer if you revert this block (the `sorted` above is fine). You'll need to add the `data = map(list, zip(*data))` back in below too.

scripts/benchmark-plot.py
```diff
@@ -1,38 +1,50 @@
...
     inputs.append(results[0]["parameters"]["input"])
 
-data = map(list, zip(*data))
-for (times, command) in zip(data, commands):
-    plt.plot(inputs, times, label=command)
+x = np.arange(len(inputs))  # the label locations
+width = 0.2  # the width of the bars
+multiplier = 0
```
philipc, line 33: Use `enumerate` for the `for` loop instead of doing it manually.

scripts/benchmark-plot.py
```diff
@@ -1,38 +1,50 @@
...
+
+fig, ax = plt.subplots(layout="constrained")
+
+for command, times in commands.items():
+    offset = width * multiplier
+    rects = ax.bar(x + offset, times, width, label=command)
+    multiplier += 1
+
+ax.set_xticks(x + width, inputs)
+ax.legend(title="Tool")
```
philipc, line 43: This can be left how it was as `plt.legend` below, since it's unrelated to the x axis.

scripts/benchmark-plot.py
```diff
@@ -1,38 +1,50 @@
...
+multiplier = 0
+
+fig, ax = plt.subplots(layout="constrained")
+
+for command, times in commands.items():
+    offset = width * multiplier
+    rects = ax.bar(x + offset, times, width, label=command)
+    multiplier += 1
+
+ax.set_xticks(x + width, inputs)
```
philipc, line 42: This isn't centered. Maybe a calculation based on `len(commands)`? Hardcode would be fine too.

philipc (changes requested):
Thanks, bars look much better.

scripts/benchmark-plot.py
philipc: (`np.transpose(data)` is probably clearer than the map/zip.)

marxin:
I addressed all your comments, they were beneficial!

Note I also formatted the code with [`ruff` tool](https://docs.astral.sh/ruff/).

scripts/benchmark-plot.py
```diff
@@ -1,37 +1,46 @@
...
     data.append([b["mean"] for b in results])
-    inputs.append(results[0]["parameters"]["input"])
 
-data = map(list, zip(*data))
-for (times, command) in zip(data, commands):
-    plt.plot(inputs, times, label=command)
+    input = results[0]["parameters"]["input"]
+    binary_size = results[0]["parameters"]["size"]
+    symbol_queries = results[0]["parameters"]["symbol_queries"]
+    inputs.append(f"{input}\n({binary_size}, {symbol_queries} qrs)")
```
philipc, line 27:
```suggestion
    inputs.append(f"{input}\n({binary_size}B,\n{symbol_queries} queries)")
```
"qrs" is a bit obscure.

philipc (approved):

PR #337, opened 2024-12-08
Cargo update for locked install

kushudai:
I cannot install addr2line with cargo with the locked flag. It works without the locked flag.

```bash
cargo install addr2line --locked
    Updating crates.io index
  Installing addr2line v0.24.2
    Updating crates.io index
error: failed to select a version for `addr2line`.
    ... required by package `backtrace v0.3.74`
    ... which satisfies dependency `backtrace = "^0.3.13"` (locked to 0.3.74) of package `addr2line v0.24.2`
versions that meet the requirements `^0.24.0` (locked to 0.24.1) are: 0.24.1

all possible versions conflict with previously selected packages.

  previously selected package `addr2line v0.24.2`

failed to select a version for `addr2line` which could resolve this conflict
```

This attempts to fix it by getting the lock file to be in sync with what cargo thinks it should use by running `cargo update` on the repo.

Fixes https://github.com/gimli-rs/addr2line/issues/338

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.15"
+version = "0.6.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
+checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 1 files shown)


kushudai:
Hi @philipc Could I trouble you to take a look at this PR?
I opened a corresponding issue as well (I apologize if that wasn't the correct etiquette, I wasn't sure)

kushudai:
> Thanks for the PR. I can reproduce this issue but I don't understand why it is happening. If I download the crate from https://crates.io/api/v1/crates/addr2line/0.24.2/download and build with `cargo build --features bin --locked` it works without error. So I doubt this PR will fix the problem for the next time we do an `addr2line` release.

Yeah, I think you are right. I was naive.
So it seems `cargo install` tries to also pull in dev-dependencies which pulls in backtrace-rs which itself depends on addr2line on non windows targets.m https://github.com/rust-lang/backtrace-rs/blob/26ab4afa46e889f2f4c81e5f5022f61569f02d16/Cargo.toml#L47
I'm confused why cargo install is pulling in dev dependencies though.

philipc:
Yeah it doesn't even build `backtrace` when you install without the `--locked`, so it shouldn't need to be checking these dependencies, and `cargo +nightly -Z avoid-dev-deps` doesn't fix the problem either. I think this is a cargo bug. We could possibly work around it by moving some tests into another crate in the workspace.

kushudai:
Thanks for your insights @philipc, I went ahead and opened an issue on cargo https://github.com/rust-lang/cargo/issues/14937

PR #342, opened 2025-02-25
fix: detecting windows root paths with unix-style separators

quanterion:
On Windows Clang can produce DWARF debug information containing paths to **DW_AT_name** and **DW_AT_comp_dir** with unix-stlye slashes. for example **C:/myproject/main.cpp**. In that case case addr2line mistakenly concatenating absolute path to file with compilation directory

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -310,5 +310,5 @@ fn has_unix_root(p: &str) -> bool {
 
 /// Check if the path in the given string has a windows style root
 fn has_windows_root(p: &str) -> bool {
-    p.starts_with('\\') || p.get(1..3) == Some(":\\")
+    p.starts_with('\\') || p.get(1..3) == Some(":\\") || p.get(1..3) == Some(":/")
 }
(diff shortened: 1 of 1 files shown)


quanterion:
@philipc can have a look at PR please?

philipc:
Seems fine but I would like to do some testing and don't currently have time.

quanterion:
ok, feel free to ping me if I can help you

philipc (changes requested):
This function is used for two purposes: to determine if a path is a root path, and to determine which separator to use. If the root path is using `:/`, then I think we also be using `/` for the separator, but that won't happen with this change. Perhaps this change should be in `has_unix_root` instead. After doing that, better names for these functions would be `has_forward_slash_root` and `has_backward_slash_root`.

philipc:
It would help with testing this if you can add a test file to https://github.com/gimli-rs/object-testfiles/tree/master/dwarf (ideally it would be the `base.cpp` in that directory compiled with clang on Windows).

quanterion:
@philipc refactored the code and here is the branch with clang binary from windows https://github.com/gimli-rs/object-testfiles/pull/22

philipc (approved):

PR #343, opened 2025-05-11
Add -j/--section option, to read offsets relative to section

noamraph:
I suggested to my coworker to use the Rust addr2line instead of the binutils, since it took a long time using the binutils version. However, he was using the `-j/--section` option of binutils addr2line, which wasn't implemented. So I implemented it. It just requires adding the section address before doing the lookup.

What do you think?

### DIFF
--- a/src/bin/addr2line.rs
+++ b/src/bin/addr2line.rs
@@ -96,6 +96,7 @@ struct Options<'a> {
     llvm: bool,
     exe: &'a PathBuf,
     sup: Option<&'a PathBuf>,
+    section: Option<&'a String>,
 }
 
 fn main() {
@@ -118,6 +119,13 @@ fn main() {
                 .value_name("filename")
                 .value_parser(clap::value_parser!(PathBuf))
                 .help("Path to supplementary object file."),
+            Arg::new("section")
+                .short('j')
+                .long("section")
+                .value_name("name")
+                .help(
+                    "Read offsets relative to the specified section instead of absolute addresses.",
+                ),
             Arg::new("all")
                 .long("all")
                 .action(ArgAction::SetTrue)
@@ -176,9 +184,15 @@ fn main() {
         llvm: matches.get_flag("llvm"),
         exe: matches.get_one::<PathBuf>("exe").unwrap(),
         sup: matches.get_one::<PathBuf>("sup"),
+        section: matches.get_one::<String>("section"),
     };
 
     let ctx = Loader::new_with_sup(opts.exe, opts.sup).unwrap();
+
+    let section_addr = opts.section.map(|section_name| {
+        ctx.get_section_address(section_name)
+            .unwrap_or_else(|| panic!("cannot find section {}", section_name))
+    });
 
     let stdin = std::io::stdin();
     let addrs = if matches.get_flag("all") {
@@ -207,6 +221,9 @@ fn main() {
                 println!();
             }
         }
+
+        // If --section is given, add the section address to probe.
+        let probe = probe.map(|x| x + section_addr.unwrap_or(0));
 
         if opts.do_functions || opts.do_inlines {
             let mut printed_anything = false;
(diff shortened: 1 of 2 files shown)


src/loader.rs
```diff
@@ -258,6 +264,7 @@ impl<'a> LoaderInternal<'a> {
 
         Ok(LoaderInternal {
             ctx,
+            object,
```
philipc, line 267: For the case of Mach-O, this may be the `object` corresponding to the `.dSYM`, whereas we probably want `--section` to be using sections from the `Mach-O`.

src/loader.rs
```diff
@@ -298,6 +305,12 @@ impl<'a> LoaderInternal<'a> {
         })
     }
 
+    fn get_section_address(&self, section_name: &str) -> Option<u64> {
+        self.object
+            .section_by_name(section_name)
```
philipc, line 310:
```suggestion
    fn get_section_address(&self, section_name: &[u8]) -> Option<u64> {
        self.object
            .section_by_name_bytes(section_name)
```

src/bin/addr2line.rs
```diff
@@ -208,6 +222,9 @@ fn main() {
             }
         }
 
+        // If --section is given, add the section address to probe.
+        let probe = probe.map(|x| x + section_addr.unwrap_or(0));
```
philipc, line 226: binutils addr2line also does a check to make sure the offset is less than the section size, and if it isn't then treats it as if no line number information was found for the address.

philipc (changes requested):
Thanks. Yes, matching binutil options is good.

src/loader.rs
noamraph: I tried to fix this in commit dac56ae31d42c24a7f1f42ab884352b923549605 - can you have a look?

src/bin/addr2line.rs
noamraph: Indeed. Fixed in 2fcede2074dbf000dace4e24bce92cfbcf1c01a2

src/bin/addr2line.rs
```diff
@@ -208,6 +222,20 @@ fn main() {
...
         }
 
+        // If --section is given, add the section address to probe.
+        let probe = probe.map(|probe| {
+            if let Some(section_range) = section_range {
+                if probe < (section_range.end - section_range.begin) {
+                    probe + section_range.begin
+                } else {
+                    // If addr >= section size, treat it as if no line number information was found.
+                    0
```
philipc, line 232: On this code path, the result should be that `probe` is assigned `None`, rather than `Some(0)`. So might need to use `and_then` instead of `map`.

src/loader.rs
```diff
@@ -209,17 +215,20 @@ impl<'a> LoaderInternal<'a> {
             })
         })() {
             let map = arena_mmap.alloc(map);
-            object = object::File::parse(&**map)?;
-        }
+            Some(object::File::parse(&**map)?)
+        } else {
+            None
+        };
+        let object1 = dsym.as_ref().unwrap_or(&object);
```
philipc, line 222:
```suggestion
        let dwarf_object = dsym.as_ref().unwrap_or(&object);
```
This looks correct now, but this might be a clearer name.

philipc (changes requested):

src/loader.rs
noamraph: Resolved in 376f95ac1f26ba097b9c47e64fe602c1eeec3116

src/bin/addr2line.rs
noamraph: Resolved in 376f95ac1f26ba097b9c47e64fe602c1eeec3116

philipc (approved):
Thanks!

noamraph:
@philipc Thank you!

Would you be able to make a new release with this change? It would be nice for my coworker if the script he writes will use a versioned release of addr2line instead of a specific commit.

philipc:
I'll do a release when I can, but it probably won't be soon sorry.

noamraph:
Sure, no problem. Thanks for letting me know, and thanks for your great
work!

PR #358, opened 2025-09-23
Use unstable instead of stable sorts

Noratrieb:
Unstable sorting has significantly better code size (and often better performance too). With this change (rebased on top of v0.25.0, which is what std currently uses) the .text size of Rust hello world (with `-Clto=fat -Copt-level=s`) goes from 297120 to 290351 (-2.23%).

I do not know *that* much about addr2line, so I cannot guarantee that this code is fine with unstable sorting. But I think it should be fine since it always sorts on addresses, which are either unique or nondeterministic anyways (as mentioned in one comment even). So it should not make a significant difference.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -158,7 +158,7 @@ impl<R: gimli::Reader> Functions<R> {
         // It's possible for multiple functions to have the same address range if the
         // compiler can detect and remove functions with identical code.  In that case
         // we'll nondeterministically return one of them.
-        addresses.sort_by_key(|x| x.range.begin);
+        addresses.sort_unstable_by_key(|x| x.range.begin);
 
         Ok(Functions {
             functions: functions.into_boxed_slice(),
@@ -252,7 +252,7 @@ impl<R: gimli::Reader> Function<R> {
         // In this example, if you want to look up address 7 at depth 0, and you
         // encounter [0..2 at depth 1], are you before or after the target range?
         // You don't know.
-        state.addresses.sort_by(|r1, r2| {
+        state.addresses.sort_unstable_by(|r1, r2| {
             if r1.call_depth < r2.call_depth {
                 Ordering::Less
             } else if r1.call_depth > r2.call_depth {
--- a/src/line.rs
+++ b/src/line.rs
@@ -94,7 +94,7 @@ impl Lines {
                 column,
             });
         }
-        sequences.sort_by_key(|x| x.start);
+        sequences.sort_unstable_by_key(|x| x.start);
 
         let mut files = Vec::new();
         let header = rows.header();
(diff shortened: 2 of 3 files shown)


philipc:
For correctness, I agree that this should only affect parts that were nondeterministic anyway. However, running `addr2line --all -af` on a `libxul.so` that I had lying around shows less addresses, so I'll try to figure out what's going on there before merging.

The performance difference when running `scripts/benchmark-addr2line-docker.sh` is noise.

`cargo bench` shows mixed results. These benchmarks aren't very stable but there does appear to be a difference here:
```
context_new             time:   [515.98 µs 516.26 µs 516.56 µs]
                        change: [+4.7649% +5.3754% +6.0487%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_new_parse_lines time:   [5.0581 ms 5.0624 ms 5.0670 ms]
                        change: [+0.1764% +0.2994% +0.4055%] (p = 0.00 < 0.05)
                        Change within noise threshold.

context_new_parse_functions
                        time:   [7.6034 ms 7.6128 ms 7.6234 ms]
                        change: [−5.9501% −5.7841% −5.6126%] (p = 0.00 < 0.05)
                        Performance has improved.

context_new_parse_inlined_functions
                        time:   [27.333 ms 27.420 ms 27.518 ms]
                        change: [+1.7455% +2.1007% +2.4944%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_query_location  time:   [832.62 µs 834.19 µs 836.49 µs]
                        change: [−3.2239% −2.8645% −2.5365%] (p = 0.00 < 0.05)
                        Performance has improved.

context_query_with_functions
                        time:   [1.9885 ms 1.9901 ms 1.9917 ms]
                        change: [+0.8561% +1.0993% +1.2944%] (p = 0.00 < 0.05)
                        Change within noise threshold.

context_new_and_query_location
                        time:   [623.46 µs 623.77 µs 624.11 µs]
                        change: [+2.3716% +3.0202% +3.5973%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_new_and_query_with_functions
                        time:   [750.20 µs 750.53 µs 750.90 µs]
                        change: [−0.2144% +0.3569% +0.8736%] (p = 0.22 > 0.05)
                        No change in performance detected.
```

Noratrieb:
I ran the benchmarks locally and also saw a few small performance slowdowns, but I'm not sure whether it's really significant. Stable vs unstable sorting performance varies a bunch, where either can be faster depending on the circumstances.

philipc (approved):
Thanks!

philipc:
> However, running addr2line --all -af on a libxul.so that I had lying around shows less addresses, so I'll try to figure out what's going on there before merging.

The `unit_ranges` sorting was one cause of this, and #359 fixes another.

